PR #28, opened 2017-11-15
Fix fname param to firstName in email templates

lilachritter:


lilachritter:
Hey,
Encountered small 'typo' bug - emails where been sent without the name in the email top row.

dortheimer:
@lilachritter , since Eyal decided that first name isn't required for registration, it doesn't exist any more. The new version I've pushed has a new email, in Hebrew and without the first name.

backend/api/service/email/newSignUp.mustache
```diff
@@ -2,7 +2,7 @@
 
 <head>
   <title>Welcome to Meirim</title>
-  <link href="https://fonts.googleapis.com/css?family=Heebo" rel="stylesheet"> </link>
+<link href="https://fonts.googleapis.com/css?family=Heebo" rel="stylesheet"> </link>
```
BlueHotDog: i think the prev indentation was correct.. if it's inside the head

backend/api/service/email/newAlert.mustache
```diff
@@ -4,7 +4,7 @@
 </head>
 <body>
 
-Hi {{fname}}!
+Hi {{firstName}}!
```
BlueHotDog: yes for full names!

PR #36, opened 2017-12-04
Scraping

BlueHotDog:
- Adding Mavat scraping
- Fixing various other issues
- Other small refactors, moving code around


backend/api/model/plan.js
```diff
@@ -1,36 +1,40 @@
...
-    if (attributes.data){
+    if (attributes.data) {
       attributes.data = JSON.stringify(attributes.data);
     }
     return super.format(attributes);
   }
+
   //
   parse(attributes) {
     try {
```
dortheimer: Empty try catch

PR #59, opened 2018-02-26
error messages from server on signup

naortor:


client/src/components/containers/home_page_form_container.jsx
```diff
@@ -74,7 +74,8 @@ class HomepageFormContainer extends Component {
 					onChange={this.onChange}
 					toggle={this.toggle}
 					terms={this.state.terms}
-					status={this.props.status}
+          status={this.props.status}
```
BlueHotDog: fix indendation :)

client/src/redux/actions/login_actions.js
```diff
@@ -2,87 +2,96 @@ import Actions from './action_types';
...
-				});
-			}, function (result) {
-				return dispatch({
-					type: Actions.SIGNUP_ERROR,
-					data: result
-				});
-			});
-		};
-	},
+  signup: function (loginForm) {
```
BlueHotDog: i think a lot of this verbose syntax can really enjoy from shorthand lambdas.. (e.g =>)

PR #143, opened 2019-06-20
add opengraph tags

florpor:
using a unified serve script which serves both static react-built assets and the api


bin/serve
```diff
@@ -0,0 +1,139 @@
...
+
+const apiService = require('../src/services/api');
+const config = require('../src/config.json');
+
+
+// create static app
+const urlencoded = BodyParser.urlencoded({
+  extended: false,
+});
+const staticApp = Express();
```
dortheimer: Let's not have two apps on two ports.
Better that we have one that serves both.

bin/serve
```diff
@@ -0,0 +1,139 @@
...
+        pageDescription = `תב"ע ב${plan.data.data.PLAN_COUNTY_NAME}`;
+
+      res.render('index', {
+        layout: false, 
+        pageTitle: `${plan.data.PL_NUMBER} - ${plan.data.PL_NAME} - מעירים`, 
+        pageDescription: pageDescription,
+        pageImage: config.opengraph.basePageImageURL,
+      });
+    }).catch(error => {
+      console.error('failed to fetch plan details for opengraph');
```
dortheimer: Use logger. Console is evil.

bin/serve
```diff
@@ -0,0 +1,139 @@
...
+      else
+        pageDescription = `תב"ע ב${plan.data.data.PLAN_COUNTY_NAME}`;
+
+      res.render('index', {
+        layout: false, 
+        pageTitle: `${plan.data.PL_NUMBER} - ${plan.data.PL_NAME} - מעירים`, 
+        pageDescription: pageDescription,
+        pageImage: config.opengraph.basePageImageURL,
+      });
+    }).catch(error => {
```
dortheimer: Use an error handler using next()

bin/serve
```diff
@@ -0,0 +1,139 @@
...
+staticApp.set('views', staticRootDir);
+staticApp.engine('html', ExpressHandlebars({extname: '.html', layoutsDir: staticRootDir}));
+staticApp.set('view engine', 'html');
+
+staticApp.get('/plan/:planId/:planName?', function(req, res) {
+  const { planId, planName } = req.params;
+
+  // fetch plan details to populate og tags
+  apiService.get('/plan/' + planId)
+    .then(plan => {
```
dortheimer: put the routes in a routes file

bin/serve
```diff
@@ -0,0 +1,139 @@
...
+        layout: false, 
+        pageTitle: 'תוכנית בניה - מעירים', 
+        pageDescription: '',
+        pageImage: config.opengraph.basePageImageURL,
+      });
+    });
+});
+
+// all non-plan pages have static og tag values
+staticApp.get('*', function(req, res) {
```
dortheimer: This doesn't make sense since we should return errors for unexisting urls.

src/config.json
```diff
@@ -2,5 +2,8 @@
   "axios": {
     "baseURL": "http://localhost:3001",
     "withCredentials": true
+  },
+  "opengraph": {
+    "basePageImageURL": "http://localhost:3000/favicon.ico"
```
dortheimer: Use the baseUrl configuration and add favicon hardcoded.

public/index.html
```diff
@@ -13,6 +13,12 @@
         <meta name="description" content="שירות המעדכן אתכם בכל פעם שמתפרסמת תוכנית בניין עיר חדשה באיזורכם" />
         <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
 
+        <meta property="og:locale" content="he_IL" />
```
dortheimer: Locale should be in the configuration.

public/index.html
```diff
@@ -13,6 +13,12 @@
         <meta name="description" content="שירות המעדכן אתכם בכל פעם שמתפרסמת תוכנית בניין עיר חדשה באיזורכם" />
```
dortheimer: The description and title should be also dynamic

package.json
```diff
@@ -93,6 +94,7 @@
   "scripts": {
     "start": "BROWSER=none ./node_modules/react-scripts/bin/react-scripts.js start",
     "server": "node bin/api",
-    "build": "./node_modules/react-scripts/bin/react-scripts.js build"
+    "build": "./node_modules/react-scripts/bin/react-scripts.js build",
+    "serve": "node bin/serve"
```
dortheimer: Should be removed.

bin/serve
florpor: done

bin/serve
florpor: we shouldn't use next() because we want to show the react 404 page

bin/serve
florpor: done

bin/serve
florpor: same as above - we shouldn't return errors for the static app so the react app loads and shows the react 404 page

src/config.json
florpor: removed

bin/serve
florpor: they are loaded into one main app

public/index.html
florpor: done

api/staticRoutes.js
```diff
@@ -0,0 +1,69 @@
...
+        layout: false, 
+        pageTitle: 'תוכנית בניה - מעירים', 
+        pageDescription: '',
+        pageImage: `${req.protocol}://${req.get('host')}/favicon.ico`,
+      });
+    });
+});
+  
+// all non-plan pages have static opengraph tag values
+Router.get('*', function(req, res, next) {
```
dortheimer: We have some specific pages, let's define them so we can respond with 404 errors when links are broken. This is very bad for seo....

api/staticRoutes.js
```diff
@@ -0,0 +1,69 @@
...
+    pageDescription: 'רוצים לדעת אם הגינה הציבורית שלכם עומדת להפוך למגדל?',
+    pageImage: `${req.protocol}://${req.get('host')}/favicon.ico`,
+  });
+});
+
+/**
+ * Convert date object to string with "dd/MM/yyyy" format.
+ */
+
+function dateStrFormat(dateStr) {
```
dortheimer: Better us arrow functions.

api/staticRoutes.js
```diff
@@ -0,0 +1,69 @@
...
+  try {
+    const dt = new Date(dateStr);
```
dortheimer: You should use moment for this.

api/staticRoutes.js
```diff
@@ -0,0 +1,69 @@
...
+  const { planId } = req.params;
+
+  // fetch plan details to populate opengraph tags
+  Plan.fetchByPlanID(planId)
+    .then(plan => {
+      let pageDescription;
+
+      // some plans don't have depositing dates
+      if (plan.attributes.data.DEPOSITING_DATE)
+        pageDescription = `תב"ע ב${plan.attributes.data.PLAN_COUNTY_NAME} אשר הופקדה ב ${dateStrFormat(plan.attributes.data.DEPOSITING_DATE)}`;
```
dortheimer: The strings should be in the template.
You can have an if condition with mustache.

api/staticRoutes.js
```diff
@@ -0,0 +1,69 @@
...
+        pageDescription: pageDescription,
+        pageImage: `${req.protocol}://${req.get('host')}/favicon.ico`,
+      });
+    }).catch(error => {
+      Log.error('failed to fetch plan details for opengraph');
+      Log.error(error);
+
+      res.render('index', {
+        layout: false, 
+        pageTitle: 'תוכנית בניה - מעירים',
```
dortheimer: Localized strings must be in the template. Just make an error template.

api/staticRoutes.js
```diff
@@ -0,0 +1,69 @@
+const Router = require('express').Router();
+
+const Log = require('./lib/log');
+const Plan = require('./model/plan');
+const config = require('../src/config.json');
+
+// all plan pages should render opengraph tags
+Router.get(['/plan/:planId', '/plan/:planId/*'], function(req, res, next) {
```
dortheimer: Better use an arrow function.

PR #148, opened 2019-07-03
Small fixes 1

dortheimer:


src/components/Wrapper/Wrapper.jsx
```diff
@@ -0,0 +1,22 @@
...
+import Navigation from "../Navigation";
+import Footer from "../Footer";
+
+export default function Wrapper(props) {
+    const { me, children } = props;
+    return (
+        <Fragment>
+            <Navigation me={me} />
+            {!children ? (
+                <div style={{ textAlign: "center" }}>
```
GalGend: why style not in class

src/scenes/About.jsx
```diff
@@ -1,118 +1,159 @@
...
+                                    </div>
+                                </div>
+                            </div>
+                            <div class="row">
+                                <div class="col-6 col-lg-4">
+                                    <div class="m-3 text-center">
+                                        <img
+                                            class="img-fluid"
+                                            src={adi}
+                                            alt="עדי קויש"
```
GalGend: why not remove them on the way

src/scenes/Home.jsx
```diff
@@ -1,61 +1,58 @@
...
-                מערכת ההתראות שלנו מאגדת במקום אחד את כל התכניות בסמכות מקומית
-                ומחוזית ותשלח לכם התראה למייל בזמן אמת, בכל פעם שתופקד תכנית
-                חדשה בסמכות מקומית או כשתקלט במערכת תכנית חדשה בסמכות מחוזית
-                באזור מגוריכם או בכל אזור אחר שתבחרו! כך תוכלו לעקוב בקלות אחרי
-                השינויים בסביבת המגורים שלכם.
-              </p>
+        <Wrapper me={me}>
+            <div
+                className="d-none d-md-block"
+                style={{
```
GalGend: why not in style

dortheimer: Since it doesn't make sense to create a call for single use.

src/scenes/About.jsx
dortheimer: @eyalmigd  can we remove adi?

src/components/Wrapper/Wrapper.jsx
dortheimer: I think for a single use there is no need to create a class in React.

PR #154, opened 2019-10-29
Ratings

dortheimer:


migrations/20190924190025_rate.js
```diff
@@ -0,0 +1,18 @@
...
+    return knex.schema.createTableIfNotExists("rate", t => {
+        t.increments("id").primary();
+        t.integer("score").notNullable();
+        t.integer("person_id").notNullable();
+        t.integer("plan_id").notNullable();
+        t.timestamp("created_at").defaultTo(knex.fn.now());
+        t.unique(['person_id', 'plan_id']);
+    }).then(()=>{
+        return knex.schema.alterTable('plan', function(t) {
+            t.string('rating', 35).notNullable().defaultTo(0);
```
boris-p: float?

src/components/Rate/Rate.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+        isLoading: true,
+        me: this.props.me || {},
+        error: false,
+        goToLogin: false,
+        max: 0,
+        numRatings: 0,
+        average: 0,
+        ratings: {
+            5: 0,
```
boris-p: i'd build this dynamically with a const for the total number

src/components/Rate/Rate.jsx
```diff
@@ -0,0 +1,167 @@
...
+        return api
+            .get("/rate/" + planId)
+            .then(d => {
+                let ratings = { 1: 0, 2: 0, 3: 0, 4: 0, 5: 0 };
+                const max = d.data.reduce(
+                    (acc, r) => (r.num > acc ? r.num : acc),
+                    0
+                );
+                d.data.map(rating => {
+                    ratings[rating.score] = rating.num;
```
boris-p: not sure what this does and if mutating is the best strategy

src/components/Rate/Rate.jsx
```diff
@@ -0,0 +1,167 @@
...
+    setRate = score => {
+        const { planId, me } = this.props;
+        if (!me) {
+            return this.setState({ goToLogin: true });
+        }
+        return api
+            .post("/rate/", { score, plan_id: planId })
+            .then(ratings => {
+                this.setState({ score });
+                this.componentDidMount();
```
boris-p: :open_mouth:

src/components/Rate/Rate.jsx
```diff
@@ -0,0 +1,167 @@
...
+            tempRate,
+            goToLogin,
+            max,
+            numRatings,
+            average,
+            ratings
+        } = this.state;
+        const displayRate = tempRate || score;
+        if (goToLogin) {
+            return <Redirect to={signInURL} />;
```
boris-p: what's the advantage of passing in an object and not a url? are we using the props passed in the redirected to component?

src/components/Rate/Rate.jsx
```diff
@@ -0,0 +1,167 @@
...
+                    <div className="col-5">
+                        <Rating
+                            direction="rtl"
+                            onClick={this.setRate}
+                            onHover={this.hoverRate}
+                            initialRating={score}
+                        />
+                    </div>
+                    <div className="col-7">
+                        {displayRate === 1 &&
```
boris-p: i'd have a const object somewhere at the top and here do something like ratingsMap[displayRate]

src/scenes/SignIn.jsx
```diff
@@ -31,7 +31,8 @@ class SignIn extends Component {
         const { me } = this.props;
 
         if (success === 1) {
-            return <Redirect to={redirectTo} />;
+            window.location = redirectTo;
+            return "";
```
boris-p: I think return is redundant here, no?

PR #163, opened 2019-12-31
add unsubscribe ability to new website

florpor:


src/scenes/AlertUnsubscribe.jsx
```diff
@@ -0,0 +1,68 @@
...
+            </Wrapper>
+        );
+    }
+
+    renderUnsubscribeStatus() {
+        const { loading, error } = this.state;
+
+        if (loading) {
+            return 'טוען...';
+        } else if (error) {
```
dortheimer: Else is not needed because you returned previously

src/scenes/AlertUnsubscribe.jsx
```diff
@@ -0,0 +1,68 @@
...
+            return 'חלה שגיאה בעת מחיקת התראה. אנא נסו שנית';
+        } else {
```
dortheimer: We don't need "else"

api/apiRoutes.js
```diff
@@ -48,6 +48,7 @@ Router.get("/alert/", wrap(Alert.browse, Alert));
 Router.get("/alert/:id", wrap(Alert.read, Alert));
 Router.post("/alert/", wrap(Alert.create, Alert));
 Router.delete("/alert/:id", wrap(Alert.delete, Alert));
+Router.delete("/alert/_token/:token", wrap(Alert.unsubscribe, Alert));
```
dortheimer: Why is there an underscore?

PR #169, opened 2020-02-04
New public service

dortheimer:


api/controller/plan.js
```diff
@@ -34,6 +38,69 @@ class PlanController extends Controller {
...
+
+        return super
+            .browse(req, {
+                columns,
+                whereRaw,
+                order,
+                pageSize:1000
+            })
+            .then(rows => {
+                rows.map(row => {
```
boris-p: would either do something like - 
response.features = rows.map(row...

or change the map function to forEach

api/controller/plan.js
```diff
@@ -34,6 +38,69 @@ class PlanController extends Controller {
...
+            Knex.raw(`ST_Intersects(geom, ST_GeomFromText("${polygon}",4326))`)
+        ];
+        const order = '-id';
+
+        return super
+            .browse(req, {
+                columns,
+                whereRaw,
+                order,
+                pageSize:1000
```
boris-p: this like a bit much for one time, don't you think?

dortheimer: This request should return all the records in a specific area, so 1000 is the top limit for a request.

PR #174, opened 2020-02-11
Setup fixes 2020

GalGend:
Hi,

I set up a new environment and updated the setup manual.
Also, fixed the migration scripts for several schemas


README.md
```diff
@@ -52,12 +52,6 @@ CREATE DATABASE meirim character set UTF8 collate utf8_bin;
 exit;
 ```
 
-Import the sql file
```
GalGend: No need to import, the migrations create the schemes

api/model/person.js
```diff
@@ -1,5 +1,5 @@
 const Promise = require('bluebird');
-const Bcrypt = require('bcrypt');
```
GalGend: BcryptJS is lighter and in pure JS

dortheimer: Is it backward compatible?
Does it require a different node version?

florpor: I'm also not sure about using this bcyptjs - the [original](https://www.npmjs.com/package/bcryptjs) package was not updated for about 3 years now, and this [@jswebfans/bcryptjs](https://www.npmjs.com/package/@jswebfans/bcryptjs) package is some fork with this single release and without a codebase (all links still refer to the original package source).

And regarding @dortheimer 's comment - maybe we should decide on the project's node version and cement it (.nvmrc or w/e method) so we are all on the same page. Currently it seems the dependencies won't install on node version 12 onwards...

dortheimer: @GalGend, can you change this back?

GalGend:
@dortheimer @florpor 
Changed back to Bcrypt

florpor:
@GalGend is this ready for merge?

GalGend:
@florpor It is

PR #178, opened 2020-02-11
Added a new module for dynamic COR

fusiongenre:
* Added a new module for dynamic COR.
 * Upgrade dependencies in order to resolve errors.


api/apiApp.js
```diff
@@ -13,13 +13,7 @@ const urlencoded = BodyParser.urlencoded({
...
 const json = BodyParser.json()
 
-// const whitelist = ['http://localhost:3000', 'http://meirim.org', 'https://meirim.org']
-const cors = Cors({
-  origin: (origin, callback) => callback(null, true),
-  optionsSuccessStatus: 200,
-  credentials: true,
-  preflightContinue: false
-})
+const cors = Cors()
```
dortheimer: This is not needed here.

api/apiRoutes.js
```diff
@@ -1,60 +1,64 @@
...
+Router.delete("/alert/:id", wrap(Alert.delete, Alert));
+Router.delete("/alert/_token/:token", wrap(Alert.unsubscribe, Alert));
 
 // me
-Router.get('/me/', wrap(Alert.browse, Alert))
+Router.get("/me/", wrap(Alert.browse, Alert));
 
 // Public API
-Router.get('/public/plan', publicWrapper(Plan.publicBrowse, Plan))
+Router.use("/public", customCors({ type: "public" }));
```
dortheimer: the variable is called Cors only

api/lib/cors.js
```diff
@@ -0,0 +1,34 @@
+const Config = require("./lib/config");
+const Cors = require("cors");
+
+module.exports = function({ type = "default" }) {
```
dortheimer: I think it's better to use lambda functions

fusiongenre:
WIP

api/lib/cors.js
```diff
@@ -0,0 +1,34 @@
...
+    const whitelist = Config.defaultAllowedOrigins;
```
dortheimer: use Config.get('defaultAllowedOrigins')

api/lib/cors.js
```diff
@@ -0,0 +1,34 @@
...
+
+    if (type === "public") {
```
dortheimer: Just export a publicCors object from this file. In this way, we don't have to pass configuration options.

config/default.json
```diff
@@ -54,5 +54,11 @@
                 "pass": ""
             }
         }
+    },
+    "defaultAllowedOrigins": ["http://localhost:3000", "http://meirim.org", "https://meirim.org"],
+    "apiTokens": {
+        "ashdfioh2io43hri2o3hdioashd": {
+            "allowedOrigins": ["https://www.simplex-mapping.com/"]
```
dortheimer: This should not be in the config file since it should be a secret. Put some placeholder instead.

PR #202, opened 2020-10-07
Added impression counter

dortheimer:
Adding impression counter and view erosion for interestingness measurements.

### DIFF
--- a/api/controller/impression.js
+++ b/api/controller/impression.js
@@ -14,7 +14,6 @@ const hashCode = (s) => {
     hash = (hash << 5) - hash + char
     hash = hash & hash // Convert to 32bit integer
   }
-  console.log(hash)
   return hash
 }
 const getIp = (req) => {
--- a/api/controller/impression.js
+++ b/api/controller/impression.js
@@ -25,12 +25,11 @@ class ImpressionController extends Controller {
     const body = req.body
     body.plan_id = parseInt(req.params.plan_id, 10)
     body.ip = hashCode(getIp(req))
-    body.person_id = req.session.person ? req.session.person.id : 0
     return (
       this.model
         .forge(body)
         .save()
-      // we don't was to ruturn data
+      // we don't want to return data
         .then(() => true)
     )
   }
--- a/api/controller/impression.js
+++ b/api/controller/impression.js
@@ -1,63 +1,84 @@
-const Bluebird = require("bluebird");
-const Controller = require("../controller/controller");
-const Impression = require("../model/impression");
-const Plan = require("../model/plan");
+const Bluebird = require('bluebird')
+const Controller = require('../controller/controller')
+const Impression = require('../model/impression')
+const Plan = require('../model/plan')
 const { Knex } = require('../service/database')
 
+const hashCode = (s) => {
+  var hash = 0
+  if (s.length === 0) {
+    return hash
+  }
+  for (var i = 0; i < s.length; i++) {
+    var char = s.charCodeAt(i)
+    hash = (hash << 5) - hash + char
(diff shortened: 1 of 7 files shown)


api/model/impression.js
```diff
@@ -0,0 +1,34 @@
...
+            ip: ["required", "string"],
+
+        };
+    }
+
+    get tableName() {
+        return "impression";
+    }
+
+    person() {
```
florpor, line 19: impression doesn't necessarily link to valid person. suggest this be removed

api/controller/impression.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Aggregate impressions and update view count in plan table.
+     * Should be executed by cron every hour or day
+     */
+    aggregate() {
+        
+        // get aggregated count for every plan
+        const query = 'select a.plan_id, count(*) as num from (SELECT plan_id, ip FROM `impression` group by plan_id, ip) as a group by a.plan_id ';
```
florpor, line 31: can avoid inner query with:
`SELECT plan_id, COUNT(DISTINCT ip) FROM impression`

api/controller/impression.js
```diff
@@ -0,0 +1,63 @@
+const Bluebird = require("bluebird");
+const Controller = require("../controller/controller");
+const Impression = require("../model/impression");
+const Plan = require("../model/plan");
+const { Knex } = require('../service/database')
+
+class ImpressionController extends Controller {
+
+    getIp = (req) => {
+        return req.headers["x-forwarded-for"] || req.connection.remoteAddress;
```
florpor, line 10: we wanted to hash the ip, no?
we don't want to keep actual client ips, plus it can save space

api/controller/impression.js
```diff
@@ -0,0 +1,63 @@
...
+                        views=views+${key}, 
+                        erosion_views=erosion_views +${key} 
+                        WHERE id IN(${map[key].join(',')})`);
+
+                // update the plan table
+                return Bluebird.mapSeries(queries, query => Knex.raw(query))
+            })
+            .then(() => {
+                // remove impressions
+                return Knex.raw('TRUNCATE impression')
```
florpor, line 57: new impressions created while update plan commands are running (can be hundreds of them) will be deleted and never counted.
do you think this is reasonable, or should a harsher mechanism be implemented?

api/model/plan.js
```diff
@@ -16,7 +18,9 @@ class Plan extends Model {
             geom: ["required", "object"],
             jurisdiction: "string",
             areaChanges: "string",
-            rating: ["required", "number"]
+            rating: ["required", "number"],
+            views: ["required", "number"],
+            erosion_views: ["required", "number"]
```
florpor, line 23: what is the meaning of this field?

src/scenes/SinglePlan.jsx
```diff
@@ -99,6 +102,7 @@ class SinglePlan extends Component {
 
         changes &&
             changes[0].forEach(change => {
+                if (!change[3]) return;
```
florpor, line 105: this shouldn't be part of this pr

florpor (changes requested):

api/controller/impression.js
dortheimer: Added a transaction. But it think it's reasonable.

src/scenes/SinglePlan.jsx
dortheimer: I bumped into a bug...

florpor: cool. the added comment is good

florpor (approved):

Issue #213, opened 2020-10-18
organize folder structure


jbesraa:
- [x]  organize folders
- [x]  make sure they run locally correctly
- [ ]  make sure tests are happy
- [x]  make sure pipeline(github actions?) is ok
- [ ]  make sure deployment is ok

PR #214, opened 2020-10-18
organize project under client and server folders

jbesraa:
#213


jbesraa:
@florpor  
* I commented out the tests just for now, because they are irrelevant, will hook up the new test setup once we push this to master(because the test setup is in other PR)
* see #213 for what I checked, the one thing I didn't tackle is the deployment , I think Ill leave that to you as you have more experience with this project deployment

florpor:
reorder looks great.
can you please remove the few .DS_Store files that slipped in?

florpor (approved):

PR #246, opened 2020-11-14
Modal style fix

nadavsofi:
In order to complete the mobile view design of email confirmation page, we have to take care of the top nav


client/src/components/Wrapper/Wrapper.jsx
```diff
@@ -1,22 +1,40 @@
...
 import Navigation from "../Navigation";
 import Footer from "../Footer";
 
 export default function Wrapper(props) {
-    const { me, children } = props;
+    const {me, children, fullPage} = props;
+    const parentRef = useRef(null);
+    const [navHeight, setNavHeight] = useState(0);
+
+    useEffect(() => {
```
jbesraa, line 12: do we need the useEffect in the wrapper?

client/src/scenes/Register/style.js
```diff
@@ -28,7 +29,7 @@ export const Title = styled.h1`
 `;
 
 export const SubTitle = styled.h2`
-  font-family: Assistant;
+  font-family: Assistant !important;
```
jbesraa, line 32: lets define a global font-family just once and not repeat it for every text

client/src/shared/dropdown/index.jsx
```diff
@@ -8,51 +8,58 @@ import styled from 'styled-components';
 import Label from '../label';
 
 const StyledSelect = styled(Select)`
+    font-family: Assistant !important;
```
jbesraa, line 11: fix indentation

jbesraa (commented):
looks good overall - there is a couple of comments

PR #252, opened 2020-11-24
Nav and header

nadavsofi:


jbesraa (approved):

client/src/components/Comments/Comment.jsx
```diff
@@ -1,30 +1,31 @@
...
-    );
+	return (
+		<li className="comment-main-level">
+			<div className="comment-avatar">
+				<img src={avatar} alt="avatar" />
+			</div>
+			<div className="comment-box">
+				<div className="comment-head">
+					<h6 className="comment-name">
+						<a href={'/profile/' + person.id}>
```
GalGend, line 16: Maybe we can remove this href, we don't use the '/profile'  page in this version

client/src/components/Comments/Comments.jsx
```diff
@@ -1,112 +1,100 @@
...
-        });
+	const handleCommentPublished = (data) => {
+		var newComment = _.merge(data, {
+			person: {
+				alias: state.me.name,
+				id: state.me.id
+			}
+		});
+		const newMe = state.me;
+		newMe.alias = newMe.alias || data.person.alias;
```
GalGend, line 43: Does the alias thing work properly? It should promote the alias input on users first comment, then save it

client/src/shared/navigation/constants.js
```diff
@@ -0,0 +1,11 @@
+import api from 'services/api'
+
+export const userLoggedInMenuItems = [
+	{
+		'text': 'התנתק',
+		'onClick': async () => {
+			const response = await api.post('/sign/out')
+			console.log('🚀 ~ file: constants.js ~ line 8 ~ \'onClick\': ~ response', response)
```
GalGend, line 8: ??

client/src/shared/navigation/desktop.jsx
```diff
@@ -0,0 +1,128 @@
...
+const DesktopNavBar = ({ user, isAuthenticated, logoutHandler }) => {
+	const dispatch = useDispatch()
+	const [dropDownEl, setDropDownEl] = React.useState(null);
+	const handleDropDownClick = (event) => {
+		setDropDownEl(event.currentTarget);
+	};
+	const handleDropDownClose = () => {
+		setDropDownEl(null);
+	};
+	const dropdownItems = [{ 'text': 'התנתק', 'onClick': logoutHandler }]
```
GalGend, line 25: Why not in the const file? or inthe t file

GalGend: There's a property exist in the HE_IL file called "signout", in Hebrew we might use the word התנתקות instead of התנתק

client/src/components/Rate/Rate.jsx
```diff
@@ -1,173 +1,301 @@
...
 
 Rate.propTypes = {
-    planId: PropTypes.number,
-    me: PropTypes.object
+	planId: PropTypes.number,
 };
 
 export default Rate;
+
+// class Rate extends Component {
```
florpor, line 153: please remove

client/src/scenes/SinglePlan.jsx
```diff
@@ -133,12 +133,12 @@ class SinglePlan extends Component {
...
 
         return (
-            <Wrapper me={me}>
+            <Wrapper>
                 {plan.PL_NAME && (
                     <div className="container">
                         <div className="container">
                             <h1>{plan.PL_NAME}</h1>
-                            {!me && (
+                            {!true && (
```
florpor, line 141: this would make this message never appear.
while this will be irrelevant in a bit with the new plan page, is there an easy way to tell here if the user is signed in or not?

client/src/shared/navigation/desktop.jsx
```diff
@@ -0,0 +1,128 @@
...
+
+DesktopNavBar.propTypes = {
+	isAuthenticated: PropTypes.bool.isRequired,
+	logoutHandler: PropTypes.func.isRequired,
+	user: PropTypes.object,
+};
+
+export default DesktopNavBar;
+
+// /* <Grid item>
```
florpor, line 121: please remove

florpor (changes requested):
these are in addition to @GalGend 's comments

client/src/scenes/SinglePlan.jsx
jbesraa: this will require migrating the page to hooks. I would pass on that atm - we are almost done with the new implemntation

florpor (approved):

PR #255, opened 2020-11-30
Tree permit

gruppin:
crawling regional tree files and save to db. 
still missing aws s3 upload


server/api/lib/trees/treesRegional/index.js
```diff
@@ -0,0 +1,3 @@
+const { regionalTreePermit } = require('../regional_tree_permit');
```
florpor, line 1: what is this file (and directory) used for?

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,210 @@
...
+	Promise.allSettled(regionalTreePermitUrls.map(url => crawlRegionalTreePermit(url)))
+		.then((results) => {
+			let sumNewPermits = 0;
+			results.forEach(element => {
+				sumNewPermits = sumNewPermits + element.value;
+			});
+			console.log(`Done! Total ${sumNewPermits} new permits`);
+		})
+		.catch(err => console.log(err))
+		.finally(() => process.exit());
```
florpor, line 205: maybe `process.exit()` should be left for the caller to use?
it is relevant when running the standalone bin file, but if we run this function from other places we might not want to exit

florpor (changes requested):

server/api/lib/log.js
```diff
@@ -1,8 +1,9 @@
+const { launch } = require('puppeteer');
```
florpor, line 1: is this here on purpose?

server/api/lib/log.js
```diff
@@ -1,8 +1,9 @@
...
 const winston = require('winston');
 
 const logger = winston.createLogger({
 	level: 'info',
-	format: winston.format.json(),
+	format: winston.format.simple(),
```
florpor, line 6: please don't push logging format changes, we rely on this in prod

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+				}
+				finally {
+					() => { clearTimeout(controllerTimeout); };
+				}
+			})();
+		});
+	}
+	catch (err) {
+		Log.error(err);
+		return Promise.resolve(); // TODO reject
```
florpor, line 82: is the todo still pending on purpose?

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+				tp.attributes[tpc.END_DATE]
+			].join('_');
+
+		if (tp.attributes[tpc.REGIONAL_OFFICE] == regionalOffice && !existingPermitsCompact.has(compact_tp)) {
+			Log.debug(`A new one! queued for saving ${compact_tp}`);
+			return tp; //original one, not compact
+		}
+	}).filter(Boolean); // remove undefined values
+	//save only the new ones
+	try { //TODO promise all or knex save bulk
```
florpor, line 129: is the todo still pending on purpose?

florpor (changes requested):

server/api/lib/trees/regional_tree_permit.js
gruppin: yes

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+
+const TreePermit = require('../../model/tree_permit');
+const tpc = require('../../model/tree_permit_constants');
+const database = require('../../service/database');
+const Config = require('../../lib/config');
+
+// Regional tree permits were taken from here: 'https://www.moag.gov.il/yhidotmisrad/forest_commissioner/rishyonot_krita/Pages/default.aspx';
+const RAW_DATA_FOLDER = require('os').homedir() + '/raw_trees';
+const regionalTreePermitUrls = [
+	'https://www.moag.gov.il/yhidotmisrad/forest_commissioner/rishyonot_krita/Documents/Befor_galil_golan.XLS',
```
GalGend, line 18: export to config file, remove spaces

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+const AbortController = require('abort-controller');
+const aws = require('aws-sdk');
+
+const TreePermit = require('../../model/tree_permit');
+const tpc = require('../../model/tree_permit_constants');
+const database = require('../../service/database');
+const Config = require('../../lib/config');
+
+// Regional tree permits were taken from here: 'https://www.moag.gov.il/yhidotmisrad/forest_commissioner/rishyonot_krita/Pages/default.aspx';
+const RAW_DATA_FOLDER = require('os').homedir() + '/raw_trees';
```
GalGend, line 16: Use 'Path' package instead, require before usage

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+	'https://www.moag.gov.il/yhidotmisrad/forest_commissioner/rishyonot_krita/Documents/Befor_darom.XLS',
+	//'https://www.moag.gov.il/yhidotmisrad/forest_commissioner/rishyonot_krita/Documents/after_darom.XLS', - no after darom
+
+];
+
+const SHEET_BEFORE = 'Data2ToExcel_BeforDate';
+const SHEET_AFTER = 'Data2ToExcel_ToDate';
+const TIMEOUT_MS = 5000;
+
+const bucketName = Config.get('aws.treeBucketName');
```
GalGend, line 42: require both const at the same line 
const {bucketName, use S3} = config.get('aws')

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+	const regionalOffice = treePermits[0].attributes[tpc.REGIONAL_OFFICE];
+	// this is the only timestamp format knex correcrtly work with 
+	const time_ago = moment().subtract(1, 'year').format('YYYY-MM-DDTHH:mm:ssZ');
+	const existingPermitsCompact = new Set();
+	await database.Knex(tpc.TREE_PERMIT_TABLE).where('updated_at', '>', time_ago.toString())
+		.andWhere(tpc.REGIONAL_OFFICE, regionalOffice)
+		.then(rows => {
+			for (let row of rows) {
+				const key_as_string =
+					[
```
GalGend, line 102: Why not use a template string instead of the join ('_')? 
Also, we can use a map function instead of the for loop

existingPermitsCompact.add(rows.map(...))

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+		const compact_tp =
+			[
+				tp.attributes[tpc.PERMIT_NUMBER],
+				tp.attributes[tpc.TREE_NAME],
+				tp.attributes[tpc.NUMBER_OF_TREES],
+				tp.attributes[tpc.END_DATE]
+			].join('_');
+
+		if (tp.attributes[tpc.REGIONAL_OFFICE] == regionalOffice && !existingPermitsCompact.has(compact_tp)) {
+			Log.debug(`A new one! queued for saving ${compact_tp}`);
```
GalGend, line 124: Log message unclear- a new what? which queue?

server/api/model/tree_permit.js
```diff
@@ -0,0 +1,41 @@
+const Model = require('./base_model');
+const tpc = require('./tree_permit_constants');
+class TreePermit extends Model {
+
+	get rules() {
+		return {
+
+			[tpc.REGIONAL_OFFICE]: 'string',
```
GalGend, line 8: all property types are string?

server/api/lib/trees/regional_tree_permit.js
```diff
@@ -0,0 +1,233 @@
...
+const parseTreesXLS = async (filename) => {
+	// hack
+	const sheetname = path.parse(filename).name.toLowerCase().includes('after') ? SHEET_AFTER : SHEET_BEFORE;
+	const workbook = xlsx.readFile(filename);
+	const sheet = workbook.Sheets[sheetname];
+	const sheet_json = xlsx.utils.sheet_to_json(sheet, { raw: false });
+	const treePermits = sheet_json.map(row => {
+		return new TreePermit(
+			{
+				[tpc.REGIONAL_OFFICE]: row['אזור'],
```
GalGend, line 150: Maybe we could save all hebrew titles at the const file, similar to the English consts. Easier when titles are changed in the excel format

server/api/model/tree_permit.js
gruppin: yes, as we don't want to suppose any assumptions on the input. if we had coordinates, it would have been of course modelled accordingly

PR #257, opened 2020-12-01
Plan

nadavsofi:


client/src/components/Wrapper/style.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+import { device } from '../../style';
+
+export const ChildrenWrapper = styled.div`
+    padding-top: 79px;
```
jbesraa, line 5: can we use something that its not px here?

jbesraa (approved):

PR #275, opened 2020-12-25
fix pdf download - solve mavat challenge

florpor:


server/api/lib/mavat/index.js
```diff
@@ -39,22 +39,99 @@ const init = () =>
...
-			response.pipe(file);
-			file.on('finish', async function () {
-				await file.close();
-				resolve(true);
-			});
+		options = options || {};
+
+		http.get(url, options, (response) => {
+			if (response.statusCode !== 200) {
+				Log.info(`downloadChallengedFile failed with status ${response.statusCode} for url ${url}`);
```
GalGend, line 48: Log error instead of info? Also, doesn't http.get goes to error if the request fails?

server/api/lib/mavat/index.js
```diff
@@ -39,22 +39,99 @@ const init = () =>
...
+				resolve(false);
```
GalGend, line 49: Why not reject instead of resolving true?

server/api/lib/mavat/index.js
```diff
@@ -39,22 +39,99 @@ const init = () =>
...
+			} else {
+				const contentType = response.headers['content-type'] || '';
+
+				// if content-type is text/html this isn't the file we wish to download but one
+				// of the challenge stages
+				if (contentType.startsWith('text/html')) {
```
GalGend, line 55: why not use `contains` rather than `startswith`

server/api/lib/mavat/index.js
```diff
@@ -39,22 +39,99 @@ const init = () =>
...
+					// if we didn't get a cookie yet this is the first part of the challenge -
+					// the page source contains the javascript code we need to run and challenge
+					// paramters for the calculation
+					if (!('set-cookie' in response.headers)) {
+						// download the entire response so we can solve the challenge
+						let responseData = '';
```
GalGend, line 61: Maybe we could extract to an external function, code too long- line 61 to 80

server/tests/unit/mavat/mavat.test.js
```diff
@@ -0,0 +1,182 @@
...
+				'<html><body>no challenge here</body></html>',
+				{'Content-Type': 'text/html'}
+			);
+
+		// create a temporary file for the result to be written to
+		const resFile = fs.createWriteStream(path.join(os.tmpdir(), `${crypto.randomBytes(8).toString('hex')}.pdf`));
+
+		// try to download the (un)challenged file
+		const resSuccess = await downloadChallengedFile(
+			'http://mavat.moin.gov.il/MavatPS/Forms/Attachment.aspx?edid=1&edn=FAKEEDN&opener=AttachmentError.aspx',
```
GalGend, line 75: extract all const to configuration, address, coockies etc not readable + not indicative

server/api/lib/mavat/index.js
florpor: no, the error event is triggered when there's a connection error, premature connection close etc.

server/api/lib/mavat/index.js
florpor: i was trying to preserve the interface and just replace the single download function

server/api/lib/mavat/index.js
florpor: because the first value in the content-type header is the mime type

server/api/lib/mavat/index.js
florpor: is the general logic clear in your opinion?

i feel like the fact that the whole challenge logic is written in order here makes it clearer than if each part was a split-off function and you had to jump up and down the source to read the logic

PR #277, opened 2020-12-28
Plan tagging

ndviv:


server/api/utils/tag_creators/base.js
```diff
@@ -0,0 +1,9 @@
+class TagCreatoerInterface {
```
florpor, line 1: typo

server/api/utils/tag_creators/hotels.js
```diff
@@ -0,0 +1,68 @@
...
+		
+		return false;
+	}
+
+	areaChangeParseNumericValue(value) {
+		if (!value.length) {
+			return undefined;
+		}
+
+		return parseFloat(value.replace("+", ""));
```
florpor, line 64: specifically change for approved state can also be negative so should handle minuses

PR #295, opened 2021-01-07
Tree permit 5

gruppin:
fe:
tree permits page
new trees alert in alert page
refactor on Alerts.js - separated to components

be:
read / write tree permit to db
enrich with geometry from nominatim (osm) - additional geocoder config
db migrations included


florpor:
closed in favour of tree permit 6 pr

PR #327, opened 2021-01-24
Issue#318

nadavsofi:


client/src/pages/Register/index.jsx
```diff
@@ -15,7 +15,8 @@ const SignupForms = () => {
 	const [firstStepSuccess, setFirstStepSucess] = useState(false);
 	const [secondStepSuccess, setSecondStepSucess] = useState(false);
 	const [firstStepValues, setFirstStepValues] = useState({ name: '', password: '', email: '' });
-	const [secondStepValues, setSecondStepValues] = useState({ type: personTypes[0].value, aboutme: '', address: '' });
```
jbesraa, line 18: why to take out the address? also, if you take out of this function you have to fix line 54 in this file

client/src/pages/Register/index.jsx
```diff
@@ -117,6 +118,7 @@ const SignupForms = () => {
 					errors={formErrors}
 					values={secondStepValues}
 					setValues={setSecondStepValues}
+					setAddress={setAddress}
```
jbesraa, line 121: you can use the setValues instead of passing setAddress and not add new state (as mentioned above)

jbesraa (changes requested):

jbesraa:
@nadavsofi please pay attention to the comments and fix merge conflict

PR #332, opened 2021-01-31
Cleanup: Fix small warnings

nonZero:


client/src/components/Mapa.jsx
```diff
@@ -55,8 +55,8 @@ const Mapa = (props) =>  {
 				geom && <GeoJSON key={geomHash} data={geom} />
 			}
 			<div className="map-title">
-				{countyName && <button className="btn btn-light disabled">{countyName}</button>}
```
florpor, line 58: this changes the way the texts in these "buttons" are styled which needs to be fixed.
plus placeholder buttons were not converted (lines 14-15)

client/src/scenes/TreePermits.jsx
```diff
@@ -124,15 +124,14 @@ class TreePermits extends Component {
 							this
 						)}
 					/>
-					<Typography variant="paragraphText" mobileVariant="paragraphText"
```
florpor, line 127: this changes the typography font.
can't say that i understand why it worked until now, but along with deleting the variants font-family should be set to Assistant

florpor (changes requested):
hey @nonZero , i finally got to start reviewing your PRs.
a couple changes here caused style changes which we won't want to deploy.
will you be able to fix those changes?

client/src/components/Mapa.jsx
nonZero: hmmm... I am not sure I understand your comment.
Is there any change in rendering between `<button>` and `<span>`?
I believe the style comes from `className="btn btn-light..."`
I think Lines 14-15 do not appear inside another `<button>` so they do not have issues.

client/src/scenes/TreePermits.jsx
nonZero: Hmm.... I don't think this code change has any effect on rendering.
There are two `Typography` components in this project - `"@material-ui/core/Typography"` and a custom one in meireim's code), is there a mixup here?

PR #336, opened 2021-02-07
add trees to homepage

gruppin:


florpor:
another solution was implemented

PR #337, opened 2021-02-08
client: remove custom eslint+prettier config + run prettier on all files

nonZero:
Hi @florpor!

I was trying to push some changes - but each file I touch makes too much diffs because of code formatting...

I highly suggest getting rid of the custom eslint+prettier fomatting rules and run prettier on all code (that's what I did in this commit).   You don't have to accept this PR - you can do it yourself.  Satisfaction guaranteed :-)

Cheers!
Udi


florpor:
yup, we should definitely sort out styling in the entire project. will take your suggestions into mind.

we will get to it sometime in the near future i hope. gonna hold off on your pr #338 until then.

PR #341, opened 2021-02-21
analytics fixes and additions

florpor:


client/config/default.json
```diff
@@ -6,5 +6,16 @@
...
         "mapsApiKey": "",
         "autocompleteDelay": 2000
+    },
+    "analytics": {
+        "ga": {
+            "enabled": false,
+            "ua": ""
+        },
+        "hotjar": {
+            "enabled": false,
```
GalGend, line 16: what are those? Maybe you can name them with clearer names....

client/src/hooks.js
```diff
@@ -52,4 +54,23 @@ export const CheckIfUserCanAccessPage = () => {
...
+			ReactGA.initialize(process.env.CONFIG.analytics.ga.ua);
+		}
+	}, []);
+};
+
+export const useInitHotjar = () => {
+	useEffect(() => {
+		if (process.env.CONFIG.analytics.hotjar.enabled) {
+			hotjar.initialize(
+				process.env.CONFIG.analytics.hotjar.hjid,
```
GalGend, line 71: please consider using clearer names for config, require them before initizializing . Also, maybe we could consider a 'config' service, that would handle getting config values from `process.env` or other resources, so it can handle it with default values, to prevent exceptions (if config not exist etc)

client/src/scenes/PrivacyPolicy.jsx
```diff
@@ -16,20 +16,20 @@ export default function PrivacyPolicy() {
                     <p style={{'font-size': '18px'}}>
                         מעירים (ע.ר.) היא המפתחת והמפעילה של הפלטפורמה, לרבות האתר ושאר השירותים המסופקים על-ידה (כולם ביחד – <b>"הפלטפורמה"</b>).
                     </p>
-                    <p>
+                    <p style={{'font-size': '18px'}}>
```
GalGend, line 19: please add `p` styles as a children of `container` (or give this an id), very repetitive (all p's that are children here get the same style of font-size:18px)

client/src/scenes/PrivacyPolicy.jsx
```diff
@@ -16,20 +16,20 @@ export default function PrivacyPolicy() {
...
                         אנו פועלים כמיטב יכולתו להגנת פרטיותם של המשתמשים, בהתאם לחוקי ותקנות הגנת הפרטיות.
                     </p>
-                    <p>
+                    <p style={{'font-size': '18px'}}>
                         אתה מוזמן לפנות אלינו בכל שאלה בנוגע למדיניות הפרטיות, לכתובת הדואר האלקטרוני: <a href="mailto:info@meirim.org">info@meirim.org</a>.
                     </p>
-                    <p>
+                    <p style={{'font-size': '18px'}}>
```
GalGend, line 32: example

client/src/scenes/PrivacyPolicy.jsx
```diff
@@ -16,20 +16,20 @@ export default function PrivacyPolicy() {
...
                     <h4><b><u>מה עניינה של מדיניות הפרטיות הזו?</u></b></h4>
                     <p style={{'font-size': '18px'}}>
                         יצרנו את מדיניות הפרטיות הזו כי אנחנו מעריכים מאוד את המידע האישי שלך; אנא קרא אותה בזהירות, שכן היא כוללת מידע חשוב בנוגע לכך.
                     </p>
-                    <p>
+                    <p style={{'font-size': '18px'}}>
                         אנו פועלים כמיטב יכולתו להגנת פרטיותם של המשתמשים, בהתאם לחוקי ותקנות הגנת הפרטיות.
                     </p>
-                    <p>
+                    <p style={{'font-size': '18px'}}>
```
GalGend, line 29: example

client/src/scenes/PrivacyPolicy.jsx
```diff
@@ -16,20 +16,20 @@ export default function PrivacyPolicy() {
...
                         האתר והפלטפורמה לא נועדו למטרות רווח והפעילות בהם הנה פעילות עמותת מעירים ולצורך קידום מטרותיה באמצעות הפעילות שלה, לרבות השימוש באתר ובפלטפורמה.
                     </p>
                     <h4><b><u>מה עניינה של מדיניות הפרטיות הזו?</u></b></h4>
                     <p style={{'font-size': '18px'}}>
                         יצרנו את מדיניות הפרטיות הזו כי אנחנו מעריכים מאוד את המידע האישי שלך; אנא קרא אותה בזהירות, שכן היא כוללת מידע חשוב בנוגע לכך.
                     </p>
-                    <p>
+                    <p style={{'font-size': '18px'}}>
```
GalGend, line 26: example

client/src/scenes/PrivacyPolicy.jsx
```diff
@@ -7,7 +7,7 @@ export default function PrivacyPolicy() {
 		<Wrapper>
             <div className="container">
                     <h2 style={{'text-align': 'center'}}><b><u>מעירים - מדיניות פרטיות</u></b></h2>
-                    <p style={{'font-size': '18px'}}><i>עדכון אחרון: 25.1.2021</i></p>
+                    <p style={{'font-size': '18px'}}><i>עדכון אחרון: 21.2.2021</i></p>
```
GalGend, line 10: example

client/src/shared/navigation/index.jsx
```diff
@@ -7,19 +8,27 @@ import MobileNavBar from './mobile';
...
+
+	useEffect(() => {
+		// tag page view
+		gaPageView(location, isAuthenticated);
+	}, [location, isAuthenticated]);
 	
 	return (
 		<React.Fragment>
-			{props.isMobile() ||props.isTablet() ?
+			{isMobile() || isTablet() ?
```
GalGend, line 31: isMobile is always defined?, what is this condition refer to? why not return true / false rather than a value and undefined?

GalGend (commented):
small comments

client/config/default.json
florpor: i assume you mean hjid and hjsv - these are the names of variables the library takes (https://github.com/abdalla/react-hotjar#readme) and also the names of variables on hotjar's tracking code info (https://help.hotjar.com/hc/en-us/articles/115011639927-What-is-the-Hotjar-Tracking-Code-), so these seem to be standard.
hjid stands for hotjar id and hjsv is hotjar service version.

client/src/hooks.js
florpor: the default.json values are the default ones and the local.json are overrides. this is the same mechanism that we are using in our server.

client/src/shared/navigation/index.jsx
florpor: yes, isMobile and isTablet are provided by the withGetScreen HOC which wraps this component. i only added props destructuring here, didn't change anything else regarding these props

client/config/default.json
GalGend: Yes I get they're hotjar conventions, however as we require them on initializaing hotjar , and they're not just loaded to `process.env` just wondered if they could be in our config more clear.

GalGend (approved):
Waiting to get some hotjar insights :)

PR #352, opened 2021-04-22
fix email wrapper support-us banner and style email code

florpor:
* change support-us banner image from svg to an attached png since gmail doesn't support svgs

* change support-us banner styling so it doesn't use flex since gmail doesn't support it

* fix code errors in email templates (extra end tags and such) and some indent and code styling issues

* add email test utility


GalGend (approved):
please attach a screenshot next time

PR #353, opened 2021-04-24
File crawling for plans

GalGend:
This PR is only for crawlign and saving mavat files


server/api/lib/mavat/files/index.js
```diff
@@ -0,0 +1,64 @@
...
+    if (file.file_icon.includes('DWG')) return FILE_TYPES.DWG;
+    return FILE_TYPES.UNKNOWN;
+}
+
+const isKML3d = ( { kind, name } ) => {
+	return kind=== 'קבצים דיגיטליים' && name.includes('KML') && name.includes('תלת מימד');
+};
+
+const getFileExtension = ( file ) => {
+    if (isKML3d(file)) return 'kml';
```
GalGend, line 20: Here I should change to use the enum

server/api/lib/mavat/index.js
```diff
@@ -349,6 +368,7 @@ const getByPlan = plan =>
 	init()
 		.then(() => {
 			return plan.get('plan_url') ? fetch(plan.get('plan_url')) : search(plan.get('PL_NUMBER'));
+			//return fetch(plan.plan_url);
```
florpor, line 371: remove comment

server/api/lib/mavat/files/index.js
```diff
@@ -0,0 +1,64 @@
+const { File } = require('../../../model');
```
florpor, line 1: file model is not exported at `api/model/index.js`. either do that or change this require

florpor (changes requested):
see comments above, and also please add a basic test for this (can be part of the existing crawler tests) so we know it works (at least basically) through package upgrades and code changes.

florpor (approved):

PR #356, opened 2021-04-26
Add windows start script

muamichali:


florpor:
hey @muamichali , there's no actual need for the existing scripts to execute files like they do now, so instead of adding `start-windows` script can you change the `start`, `build` and `cypress` scripts to execute `react-app-rewired` and `cypress` the way you did in the new script you added?
that way we won't have different scripts for different os's

PR #358, opened 2021-04-27
Client: Made WeNeedYou Box Clickable

### LINKED ISSUE #348: make entire "we need you" banner clickable
right now only the button inside the banner is clickable and links to the support us page.
change it so that clicking anywhere in the banner leads there.

omriperi:
Redirecting to "Support Us" page

Fixes #348

### DIFF
--- a/client/src/shared/weNeedYou/index.jsx
+++ b/client/src/shared/weNeedYou/index.jsx
@@ -13,6 +13,7 @@ const Box = styled.div`
 	width: 100%;
 	flex-direction: column;
 	padding-top: 14px;
+	cursor: pointer;
 
 	@media ${device.tablet} {
 		width: 1236px;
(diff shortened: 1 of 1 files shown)


client/src/shared/weNeedYou/index.jsx
```diff
@@ -93,7 +93,7 @@ const WeNeedYou = () => {
 
 	return (
 		<WeNeedYouSection>
```
florpor, line 95: the entire section can hold the onClick event (also please remove it from the button so we don't accidentally mess this up if we ever want to change this).
also please add `cursor: pointer` css rule so the whole thing behaves as if it was a link/button.

florpor (changes requested):

florpor (approved):

Issue #369, opened 2021-05-22
Tags DB


GalGend:
Hi,

This task stands for creating tag tables in DB. It is required before backend functionality of tags. See proposed design [here](https://docs.google.com/presentation/d/1zlAGXRPWSy6C1LjmBDCl1IwKa8953pt2DwBQ5-93XAw/edit?usp=sharing)

- [x] Create a tag DB table migration, to include the following fields:
id | integeer
parent_id | integer, Key to tags from this table
display_name | text
score | Integer, default 0
is_super_tag | Boolean
is_stand_alone | Boolean
display_tooltip | text

- [x] Create a tag DB model with the same fields

- [x] Create a DB plan_tag model, that would represent a plan's tags. Make sure it populates a plan (association table).
Include the following fields:
id | integer
tag_id | FK to tags
plan_id | FK to plans
display_score | Integer, default 1
created_by_ data_rules | Text in list/json format
created_by_child | Boolean, default FALSE
child_ss_standAlone | Boolean, default&nbsp; False
creation_date | date/time

- [x] Add tags as a field of plans, populating it from this table

GalGend:
Hi @TomerMe2 @muamichali Please update status of this task

muamichali:
I think we can close this issue. Everything here is done. 

But please note that the crawler still needs to call this code.

PR #389, opened 2021-06-22
add tag tables, tag housing & public related plans

muamichali:
starts issue #369 

do not merge until #403 is done

### DIFF
--- a/server/api/constants.js
+++ b/server/api/constants.js
@@ -14,16 +14,35 @@ exports.personTypes = {
 	'6': 'אחר'
 };
 
-exports.tagDataRules = {
-	housingByArea: {
-		usage: 'מגורים (מ"ר)' ,
-		minValue: 1000,
-		description: 'adds more than 1,000 Sq Meters of housing'
+exports.tagDataRules = 
+[
+	{
+	tagId: 1,
+	tagName: 'דיור',
+	rules: [
+		 {
+			usage: 'מגורים (מ"ר)' ,
+			minValue: 1000,
+			description: 'adds more than 1,000 Sq Meters of housing'
+		},
+		 {
+			usage: 'מגורים (יח"ד)',
+			minValue: 10,
+			description: 'adds more than 10 units of housing'
+		}]
 	},
-	housingByUnits: {
-		usage: 'מגורים (יח"ד)',
-		minValue: 10,
-		description: 'adds more than 10 units of housing'
-	}
-	
+	{ 
+	tagId: 3,
+	tagName: 'מבני ציבור',
+	rules: [
+		{
+			usage: 'מבני ציבור (מ"ר)' ,
+			minValue: 50,
+			description: 'adds more than 50 Sq Meters of public area'
+		}]
 }
+
+]
+	
+	
+	
--- a/server/api/constants.js
+++ b/server/api/constants.js
@@ -14,9 +14,9 @@ exports.personTypes = {
 	'6': 'אחר'
 };
 
-exports.tags = [{
-	'tagName':'דיור' , 'tagId':'1', 'tagHelperFunction':'isHousing ({{{planId}}},{{{tagId}}})'
-}]
+exports.tags = [
+	{'tagName':'דיור' , 'tagId':'1', 'functionName': 'isHousing'}
+]
 
 exports.tagDataRules = {
 	housingByArea: {
--- a/server/api/constants.js
+++ b/server/api/constants.js
@@ -13,3 +13,7 @@ exports.personTypes = {
 	'5': 'יזם',
 	'6': 'אחר'
 };
+
+exports.tags = {
+	'דיור': '1'
+}
--- a/server/api/constants.js
+++ b/server/api/constants.js
@@ -22,12 +22,12 @@ exports.tagDataRules = {
 	housingByArea: {
 		usage: 'מגורים (מ"ר)' ,
 		minValue: 1000,
-		description: 'adds more than 1,000 square meters of housing'
+		description: 'adds more than 1,000 Sq Meters of housing'
 	},
 	housingByUnits: {
 		usage: 'מגורים (יח"ד)',
 		minValue: 10,
 		description: 'adds more than 10 units of housing'
-	}	
+	}
 	
 }
--- a/server/api/constants.js
+++ b/server/api/constants.js
@@ -17,3 +17,17 @@ exports.personTypes = {
 exports.tags = {
 	'דיור': '1'
 }
+
+exports.tagDataRules = {
+	housingByArea: {
+		usage: 'מגורים (מ"ר)' ,
+		minValue: 1000,
+		description: 'adds more than 1,000 square meters of housing'
+	},
+	housingByUnits: {
+		usage: 'מגורים (יח"ד)',
+		minValue: 10,
+		description: 'adds more than 10 units of housing'
+	}	
+	
+}
--- a/server/api/constants.js
+++ b/server/api/constants.js
@@ -14,10 +14,6 @@ exports.personTypes = {
 	'6': 'אחר'
 };
 
-exports.tags = [
-	{'tagName':'דיור' , 'tagId':'1', 'functionName': 'isHousing'}
-]
-
 exports.tagDataRules = {
 	housingByArea: {
 		usage: 'מגורים (מ"ר)' ,
--- a/server/api/constants.js
+++ b/server/api/constants.js
@@ -14,9 +14,9 @@ exports.personTypes = {
 	'6': 'אחר'
 };
 
-exports.tags = {
-	'דיור': '1'
-}
+exports.tags = [{
+	'tagName':'דיור' , 'tagId':'1', 'tagHelperFunction':'isHousing ({{{planId}}},{{{tagId}}})'
+}]
 
 exports.tagDataRules = {
 	housingByArea: {
--- a/server/api/controller/cron.js
+++ b/server/api/controller/cron.js
@@ -213,21 +213,24 @@ const sendTreeAlerts = () => {
 
 const updatePlanTags = async () => {
 	Log.info('Re-creating plan tags');
+	let tagCounter = 0;
 	// Re-compute the tags of a plan if the last update time of the plan is after the last update time of the tags of this plan.
 	// Before re-computing the tags of a plan, remove all previous tags for this plan.
 	
 	// TODO: Loop on the plans that need to be updated , looping on all plans for now because none have tags 
 	// TODO: Refactor to async await style
-	const plans = await Plan.getPlansToTag({limit: 2000});
-	console.log(`I found ${plans.models.length} plans, friendo`);
+	const plans = await Plan.getPlansToTag();
+	Log.info(`Processing ${plans.models.length} plans`);
 	for (const plan in plans.models) {
 		const planId = plans.models[plan].get('id');
 		await PlanTag.deletePlanTags(planId);
 		const tags = await PlanTag.generateTagsForPlan(planId);
-		if (tags){
+		if (tags && tags.length>0){
 			await PlanTag.createPlanTags( tags);
+			tagCounter++;
 		}
 	}
+	Log.info(`Done. Added tags to ${tagCounter} plans.`);
 };
 
 /** Private */
--- a/server/api/controller/cron.js
+++ b/server/api/controller/cron.js
@@ -222,8 +222,8 @@ const updatePlanTags = async () => {
 		const planId = plans.models[plan].get('id');
 		await PlanTag.deletePlanTags(planId);
 		const tags = await generateTagsForPlan(planId);
-		if (tags && tags.length>0){
-			await PlanTag.createPlanTags( tags);
+		if (tags && tags.length > 0){
+			await PlanTag.createPlanTags(tags);
 			tagCounter++;
 		}
 	}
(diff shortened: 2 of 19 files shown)


muamichali (commented):
- [x] refactor from bluebird to async await
- [x] change save function to accept multiple plan-tag thingies
- [ ] it's ok to implement in sql as long as it's readable
- [ ]  add tests for all code
- [ ] update error handling to use the project's Exception class 
- [ ] show to @TomerMe2 after implementing housing

server/api/controller/cron.js
```diff
@@ -210,6 +208,29 @@ const sendTreeAlerts = () => {
...
+	// Before re-computing the tags of a plan, remove all previous tags for this plan.
+	
+	// TODO: Loop on the plans that need to be updated 
+	const plans = await Plan.getPlansToTag();
+	Log.info(`Processing ${plans.models.length} plans`);
+	for (const plan in plans.models) {
+		const planId = plans.models[plan].get('id');
+		await PlanTag.deletePlanTags(planId);
+		const tags = await generateTagsForPlan(planId);
+		if (tags && tags.length>0){
```
TomerMe2, line 225: spaces around >

server/api/controller/cron.js
```diff
@@ -210,6 +208,29 @@ const sendTreeAlerts = () => {
...
+			await PlanTag.createPlanTags( tags);
```
TomerMe2, line 226: remove space before tags

server/api/controller/plan_tag.js
```diff
@@ -0,0 +1,20 @@
+const Controller = require('../controller/controller');
+const PlanTag = require('../model/plan_tag');
+const Log = require('../lib/log');
+
+
+class PlanTagController extends Controller {
+	/**
+	 * Return person's alerts. Must be logged in.
```
TomerMe2, line 8: I think that this comment is not related to this class

server/api/lib/tags/index.js
```diff
@@ -0,0 +1,62 @@
+ const { tags, tagDataRules } = require('../../constants');
+ const PlanAreaChanges = require('../../../api/model/plan_area_changes');
+
+ const  generateTagsForPlan = async (planId) => {
+    let planTags = [];
```
TomerMe2, line 5: please change this let to const

server/api/lib/tags/index.js
```diff
@@ -0,0 +1,62 @@
...
+    let dataRules = [];
```
TomerMe2, line 6: please change this let to const

server/api/lib/tags/index.js
```diff
@@ -0,0 +1,62 @@
...
+    planTags.push( {
+        plan_id: planId,
+        tag_id: housingTag,
+        display_score: 0, /* TODO: Add the correct display score here */
+        created_by_data_rules: `[${dataRules.toString(',')}]`
+    });
+    return planTags;
+}
+
+const isTag = async (planId, tag, rule) => {
```
TomerMe2, line 33: Can you move this function into housing_tag.js os something similliar?
After discussing with the team, it's best best to keep every tag in a seperate file.

server/api/lib/tags/index.js
```diff
@@ -0,0 +1,62 @@
+ const { tags, tagDataRules } = require('../../constants');
+ const PlanAreaChanges = require('../../../api/model/plan_area_changes');
+
+ const  generateTagsForPlan = async (planId) => {
+    let planTags = [];
+    let dataRules = [];
+    // HOUSING
+    const housingTag = tags['דיור'];
```
TomerMe2, line 8: Can you iterate on all the keys of the dictionary of tags instead of using just the housing key?
We want to make this function to work automatically on new tags that will arrive without adding code to this function. We would like this function to be 0 code maintenance for as long as we can.

I think that combining with moving the isTag function to individual files (a file per tag), you can move the function of "how to handle this dict key" to this file as well. (see the comment about moving isTag to seperate files below)

server/api/lib/tags/index.js
```diff
@@ -0,0 +1,62 @@
...
+        tag_id: housingTag,
+        display_score: 0, /* TODO: Add the correct display score here */
+        created_by_data_rules: `[${dataRules.toString(',')}]`
+    });
+    return planTags;
+}
+
+const isTag = async (planId, tag, rule) => {
+    try {
+        const result = await PlanAreaChanges.byPlanAndUsage(planId, tagDataRules[rule].usage);
```
TomerMe2, line 35: Please add spaces around commans or operators in this function

server/api/model/plan_area_changes.js
```diff
@@ -0,0 +1,55 @@
+const Model = require('./base_model');
+const Log = require('../lib/log');
+const Exception = require('./exception');
+const { Bookshelf, Knex } = require('../service/database');
+const {	tags, tagDataRules } = require('../constants');
+
+class PlanAreaChanges extends Model {
```
TomerMe2, line 7: After the merge of the other PR, you sohuld merge it to this branch. I put some comments about this file on the other PR.

TomerMe2 (changes requested):
Great work! Thank you!
We should make the process of creating the tags more generic in my opinion. It will be better because we will have to write less code when introducing new tags and it will be easier to implement new tags.

server/api/lib/tags/index.js
muamichali: but I add things to it, so it's not a const...

server/api/lib/tags/index.js
muamichali: but I add things to it, so it's not a const...

server/api/lib/tags/index.js
TomerMe2: Cost and let are referring to the reference. The reference is set once at your code, on deceleration time.
After that you are mutating the object, but don't change the reference. In that scenario, const is preferable. 

Legit code using const:
```
const a = [];
a.push(2);  //mutating the object
```
Non legit code that will fail:
```
const a = [];
a = [2];   // mutate the reference
```

server/api/model/plan_area_changes.js
muamichali: Done.

server/api/lib/tags/housing.js
```diff
@@ -0,0 +1,36 @@
...
+const { isTagByUsageAddition} = require('../tags/utils');
+
+const isHousing = async (planId, tag) => {  
+	const HOUSING_BY_AREA = 'housingByArea';
+	const HOUSING_BY_UNITS = 'housingByUnits';
+	const planTags = [];
+	const dataRules = [];
+	
+	// check housing by area
+	const housingByArea = await isTagByUsageAddition(planId,tag,HOUSING_BY_AREA);
```
TomerMe2, line 11: Please add spaces between operands

server/api/lib/tags/housing.js
```diff
@@ -0,0 +1,36 @@
...
+	
+	// check housing by area
+	const housingByArea = await isTagByUsageAddition(planId,tag,HOUSING_BY_AREA); 
+	if (housingByArea) {
+		dataRules.push(
+			housingByArea.created_by_data_rules
+		);
+	}
+	// check housing by units
+	const housingByUnits = await isTagByUsageAddition(planId,tag,HOUSING_BY_UNITS);
```
TomerMe2, line 18: Please add spaces between operands

server/api/lib/tags/utils.js
```diff
@@ -0,0 +1,33 @@
+const { tagDataRules } = require('../../constants');
+const PlanAreaChanges = require('../../../api/model/plan_area_changes');
+
+const isTagByUsageAddition = async (planId, tag, rule) => {
+	try {
+		const result = await PlanAreaChanges.byPlanAndUsage(planId, tagDataRules[rule].usage);
+		if (result && result.models && result.models[0]) {
+			const changeToApprovedState = result.models[0].attributes.change_to_approved_state;
+			if ( (changeToApprovedState.length>1) &&
```
TomerMe2, line 9: please add spaces between operators and operands

TomerMe2 (changes requested):
Great work! Thanks!

server/api/constants.js
```diff
@@ -13,3 +13,21 @@ exports.personTypes = {
 	'5': 'יזם',
 	'6': 'אחר'
 };
+
+exports.tags = [
+	{'tagName':'דיור' , 'tagId':'1', 'functionName': 'isHousing'}
```
GalGend, line 18: When using objects, no need to add `'` for keys, no need to change but for next tine this can also work:
```
{tagName:'housing'....}
```

server/api/lib/tags/housing.js
```diff
@@ -0,0 +1,36 @@
...
+		planTags.push( {
+			plan_id: planId,
+			tag_id: tag,
+			display_score: 0, /* TODO: Add the correct display score here */
+			created_by_data_rules: `[${dataRules.toString(',')}]`
+		});
+	return planTags;
+}
+
+module.exports = {
```
GalGend, line 34: General structure for tags:

We have a folder of tags= housing, employment, commerce, etc. 
Each file (housing.js) should export one function- `doesApply` or `doesTagApply` or any name you find suitable. Then, we have an index.js file with all tags, their label (as done in constants.js, but in our tag folder) and the isApply function.

GalGend: Then, the code that will run the tag-extraction can access all rules using the index.js file, and shouldn't know it explicitly

server/api/controller/cron.js
```diff
@@ -210,6 +208,29 @@ const sendTreeAlerts = () => {
...
 
+const updatePlanTags = async () => {
+	let start = Number(Date.now());
+	Log.info('Re-creating plan tags');
+	let tagCounter = 0;
+	// Re-compute the tags of a plan if the last update time of the plan is after the last update time of the tags of this plan.
+	// Before re-computing the tags of a plan, remove all previous tags for this plan.
+	
+	// TODO: Loop on the plans that need to be updated 
+	const plans = await Plan.getPlansToTag();
```
GalGend, line 219: Do we set limits on how many plans we fetch to extract tags from?

GalGend (commented):
Hi, looks good :) Left a general comment on how we should structure our tags and rules files.

server/api/lib/tags/housing.js
muamichali: Sounds good. I refactored it based on this feedback in https://github.com/meirim-org/meirim/pull/389/commits/7583ef1a2cfe5292725020c9e3e41870d8323ba8 PLMK if it looks good!

server/api/controller/cron.js
muamichali: After discussing with @TomerMe2 I am going to implement the following logic in getPlansToTag: 
1. add an indicator of when we last updated the tags to the plan table
2. update tags last updated indicator when we update tags
3. compare it to when the plan was last updated in this function. 

Do you want to also set a max number of records?

GalGend: Hi Michal
As for now, Tomer explained to me that the tag generation is an easy task, and doesn't require loads of processing. Anyways, I think it would be better if we do manage the number of plans we generate tags for, as our server has no resources

server/api/controller/plan_tag.js
```diff
@@ -0,0 +1,20 @@
...
+
+
+class PlanTagController extends Controller {
+	/**
+	 * Return tags for this plan.
+	 * @param {IncomingRequest} req
+	 */
+	byPlan (req) {
+		return this.model.byPlan(req.params.plan_id).then(collection => {
+			Log.debug(this.tableName, 'Get tag list', req.params.plan_id);
```
GalGend, line 13: What fields of the tag it returns? Its ID or populate it with the tag table to get its labels?

server/api/lib/tags/utils.js
```diff
@@ -0,0 +1,33 @@
+const { tagDataRules } = require('../../constants');
+const PlanAreaChanges = require('../../../api/model/plan_area_changes');
+
+const isTagByUsageAddition = async (planId, tag, rule) => {
+	try {
+		const result = await PlanAreaChanges.byPlanAndUsage(planId, tagDataRules[rule].usage);
+		if (result && result.models && result.models[0]) {
```
GalGend, line 7: Also, I think it'd be better if the rule description should be done in the housing file, and the isTagByUsageAddition function should also return true/ false.

For example, send to the function `isTagByUsageAddition` what it needs to check, and it'll return true/ false. Then the housing will add it .

In this case, the `isTagByUsageAddition` function does not know the logic of the rules, and only performs what it's meant to do - check a specific land use change. Also, it will not require sending it the tag info

```
if (housingByUnits) {
		dataRules.push(
			created_by_data_rules: {   
						tag_id: tag, 
						created_by_data_rules: `{rule:'${tagDataRules[rule].description}',detail:'adds ${changeToApprovedState} ${tagDataRules[rule].usage}'}`
					}
		);
	}

```

server/api/controller/plan_tag.js
muamichali: It returns the tag name and tag display name from the Tag table. See https://github.com/muamichali/meirim/blob/7583ef1a2cfe5292725020c9e3e41870d8323ba8/server/api/model/plan_tag.js#L37

server/api/controller/cron.js
muamichali: CC @TomerMe2 when you come back. I am talking to @GalGend and she says that we don't handle plan updates yet, so maybe this is not necessary yet?

TomerMe2 (approved):

PR #413, opened 2021-08-10
Plan Area Changes In Crawler & Migration

TomerMe2:
This PR is for adding plan area changes as a migration to be run on the server and for updating the crawler to add plan area changes as part of the crawling process.

### DIFF
--- a/server/bin/populate_plan_area_changes.js
+++ b/server/bin/populate_plan_area_changes.js
@@ -7,14 +7,14 @@ const populatePlanAreaChanges = async () => {
 			qb.where('areaChanges', '!=', '[[]]').select('id','areaChanges');
 		})
 		.fetchAll();
-		areaChanges = result.models;
+		const areaChanges = result.models;
 		let howMuchLeft = areaChanges.length;
 		for (let counter = 0; counter < areaChanges.length; counter++) {
 			const planId = areaChanges[counter].attributes.id;
 			const rawAreaChangesString = areaChanges[counter].attributes.areaChanges;
 			console.log(`${howMuchLeft} plans left`);
 			howMuchLeft--;
-			await PlanAreaChangesController.refreshPlanAreaChanges(planId, rawAreaChangesString)
+			await PlanAreaChangesController.refreshPlanAreaChanges(planId, rawAreaChangesString);
 			
 		}		
 	}
--- a/server/bin/populate_plan_area_changes.js
+++ b/server/bin/populate_plan_area_changes.js
@@ -9,14 +9,22 @@ const populatePlanAreaChanges = async () => {
 		.fetchAll();
 		const areaChanges = result.models;
 		let howMuchLeft = areaChanges.length;
+
+		console.log('starting to populate plan_area_changes table');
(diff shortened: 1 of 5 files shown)


server/migrations/20210810194844_populate_area_changes.js
```diff
@@ -0,0 +1,10 @@
+const populatePlanAreaChanges = require('../bin/populate_plan_area_changes');
+
+
+exports.up = async function(knex) {
+  await populatePlanAreaChanges();
+};
+
+exports.down = function(knex) {
+  knex.raw('DELETE FROM plan_area_changes');
```
GalGend, line 9: Can you please use 
```
 await knex.schema.dropTableIfExists('plan_area_changes');
```

server/bin/populate_plan_area_changes.js
```diff
@@ -7,21 +7,36 @@ const populatePlanAreaChanges = async () => {
...
 	catch (err) {
 		console.log(err);
 	}
 
 };
 
-populatePlanAreaChanges().then(() => console.log('done populating plan area changes'));
+
+
+if (require.main === module) {
```
GalGend, line 37: What is this condition for?

GalGend (approved):

server/bin/populate_plan_area_changes.js
TomerMe2: This file is intended to be a runnable file and an importable file. This if is checking if this file ia a main. If it's a main, we should execute the function. If it's loaded as an import, we shouldn't execute the function.

server/migrations/20210810194844_populate_area_changes.js
TomerMe2: It shouldn't drop the table, it should remove all the entries. I changed this line to be more Javascript-y and that will delete all the rows.

server/bin/populate_plan_area_changes.js
GalGend: Why don't we separate it into 2?

server/migrations/20210810194844_populate_area_changes.js
GalGend: Where's this table is created from? When it is created?

server/bin/populate_plan_area_changes.js
TomerMe2: Why? It's the same code.

server/migrations/20210810194844_populate_area_changes.js
TomerMe2: One migration before this one. It is created in a previously accepted PR.

GalGend: oh ok then

server/bin/populate_plan_area_changes.js
GalGend: But two different purposes? The runnable (executable) file will *only * call the 1 function in the logic file

TomerMe2: Fixed by splitting this file into a code file and a runnable, and put them in a folder inside /bin.

Issue #395, opened 2021-07-13
New Plan Card Design


drorgivton:
This is a UI task to implement our new design for the plan card.

It should include the following information:
![image](https://user-images.githubusercontent.com/66066681/125492322-e165f585-d351-47fb-aa54-17ff4877c5b4.png)

[Link to Figma](https://www.figma.com/file/9YXElmI7lgM40Z507LBP6E/Meirim-(Tal's-file)?node-id=5666%3A83614)

- [ ] simplified status + date. The circle next to the status has a dynamic color based the status. if "מאושרת" then blue, else purple. See #372 
- [ ] button to bookmark and cancel bookmarking of the plan 
- [ ] map
- [ ] city name
- [ ] area of plan (דונם)
- [ ] distance from search coordinates. this is a dynamic element: if the context of the card is not after an address search, hide this element (just like the existing behavior).
- [ ] difference of housing units (דירות). if added ,the add "+". if removed, add "-" after the number. If no changes, hide this element.
- [ ] plan name #377 
- [ ] plan tags. #369 read from the plan_tag table per plan.
- [ ] alert (in orange) - no need to develop now. 

**Additional Requirements**
- [ ] Card should have a fixed height. If the tags exceed 2 lines, show only the results that can be visible, and add (X) showing the number of tags that couldn't be shown.
- [ ] Card width is dynamic: In mobile mode, width is 270px, while the in tablet/desktop it is 360px

drorgivton:
Great Job @ofirkg this is moved to Closed

PR #414, opened 2021-08-22
New plan card design

ofirkg:
#395 

added the new plan card design + some server fixes to accommodate new client design.


client/src/shared/planCard/index.jsx
```diff
@@ -7,53 +7,340 @@ import { Grid } from '@material-ui/core';
 import { useTheme } from '@material-ui/core/styles';
 import UnsafeRender from 'components/UnsafeRender';
 import { Text } from 'shared';
+import moment from 'moment'
+import styled from 'styled-components';
+import BookmarkOutlinedIcon from '../../assets/bookmark-outlined-icon.svg';
```
GalGend, line 12: Can you please create an index file for all the icons? So we import them all at once?

client/src/shared/planCard/index.jsx
```diff
@@ -7,53 +7,340 @@ import { Grid } from '@material-ui/core';
...
 const PlanCard = ({ plan }) => {
 	const theme = useTheme();
+	const tagsWrapperRef = useRef(null);
+	const [tags, setTags] = useState(plan?.tags || []);
+    const [ isFavPlan, setIsFavPlan ] = useState(false);
+	const tagsRef = useRef([...(plan?.tags || []).map(() => createRef()), createRef()]);
+    const dispatch = useDispatch();
+    const { isAuthenticated, user } = UserSelectors();
+
+    useEffect(() => {
```
GalGend, line 31: We can store liked plans in the store, then use selectors to get the ids of the liked plans, instead of fetching them

client/src/shared/planCard/index.jsx
```diff
@@ -7,53 +7,340 @@ import { Grid } from '@material-ui/core';
...
+        const handler = async () => {
+            await getIsFav();
+        };
+        handler();
+    }, [getIsFav]);
```
GalGend, line 36: Also, can we please extract the SubscriptionHandler to another FC? I mean, it will represent the subscription status for a plan per person, and will render its status and handle like/unlike.

In this case, we can share code between plan page and plan card
Something like that:

```
export const PlanSubscriptionStatus = ({ planId }) => {
// Getting 'liked' status from store/ server etc
const isSubscribed = useSelector(.....);
// here rendering the bookmark- full or not
return <bookmarkIcon filled={isSubscribed} tooltipText={isSubscribed:'unsubscribe': "subscribe"})
}
```

client/src/shared/planCard/index.jsx
```diff
@@ -7,53 +7,340 @@ import { Grid } from '@material-ui/core';
...
-							/>
-						</SC.CardMedia>
-						<SC.CardContent>
+					<SC.CardMedia title={plan.PL_NUMBER}>
+						<MapTitle>
+							<StatusChip>
+								<StatusDot approved={plan.status === 'מאושרות'} />
+
+								<Text
+									size='14px'
```
GalGend, line 131: Can this be styled component? instead of declaring styles in the HTML?

client/src/shared/planCard/index.jsx
```diff
@@ -7,53 +7,340 @@ import { Grid } from '@material-ui/core';
...
+    background: #F5F5F5;
+    border-radius: 4px;
+    padding: 0 8px;
+    margin: 0 0 10px 10px;
+    
+    ${({ type }) => type && `
+      &:before {
+        width: 20px;
+        height: 20px;
+        // content: url(${require('../../assets/'+type+'-tag-icon.svg')});
```
GalGend, line 337: Please use the index file for the requires

GalGend (commented):
Nice Job Ofir, can't wait to see it in our app.

Added some comments

Thanks!

client/src/shared/planCard/index.jsx
ofirkg: done 14529a8a61e5337e4846adf586386e4e1bc3ae0c

client/src/shared/planCard/index.jsx
ofirkg: done 14529a8a61e5337e4846adf586386e4e1bc3ae0c

Issue #415, opened 2021-08-24
Stream (נחל) Tag


TomerMe2:
Table 4 of this plan contains father_category that contains the term: "נחל". (contains and not equals)

Python code for this tag is attached.
[stream.zip](https://github.com/meirim-org/meirim/files/7039863/stream.zip)

PR #422, opened 2021-08-29
[WIP] Forest and River Tags

### LINKED ISSUE #416: Forest (יער) Tag
Table 4 of this plan contains father_category that contains the term: "יער". (contains and not equals)

Python code for this tag is attached.
[forest.zip](https://github.com/meirim-org/meirim/files/7039908/forest.zip)

TomerMe2:
Closes #416 #415


server/api/lib/tags/index.js
```diff
@@ -1,25 +1,39 @@
...
+
+const getPlanTagger = async () => {
+	const resources = await getTagsResources();
+
+	return async (plan) => {
+		const planTags = [];
+		for (const tagFunction of taggingFunctions) {
+			const result = await tagFunction.doesTagApply(plan, resources);
+
+			if (result !== null) {
```
GalGend, line 27: the condition can be written like this:
```
if(result) planTags.push(result)
```

server/api/lib/tags/index.js
```diff
@@ -1,25 +1,39 @@
...
-	return planTags;	
-}
+		return planTags;
+	};
+};
+
 
 module.exports = {
-	generateTagsForPlan
+	getPlanTagger
```
GalGend, line 38: Prefereably renamed to `PlanTagger` or without a name, by using the following syntax
```
module.exports  = PlanTagger;
```

Then, when requiring it, it can be given any name:

```
const PlanTagger = require('../lib/tags');
const Tagger = require('../lib/tags'); // or any other name
```

server/api/lib/tags/light_rail.js
```diff
@@ -0,0 +1,32 @@
...
+        if (planName.includes('רכבת קלה')) {
+            return getReturnDict('רכבת קלה');
+        }
+        if (planName.includes('רק"ל')) {
+            return getReturnDict('רק"ל');
+        }
+
+    }
+
+    return null;
```
GalGend, line 26: No need to return null, it's the default, always returned

server/api/lib/tags/light_rail.js
```diff
@@ -0,0 +1,32 @@
...
+          display_score: 0, /* TODO: Add the correct display score here */
+          created_by_data_rules: JSON.stringify([{ rule: `has "${stringInPlanName}" in plan name` }])
+      };
+    };
+
+    const planName = plan.attributes.PL_NAME;
+
+    if (!planName.includes('צומת') && !planName.includes('ציר')) {
+
+        if (planName.includes('רכבת קלה')) {
```
GalGend, line 17: please export to consts at top of file
```
const LightRailName = 'רקל'
const tzomet = 'צומת'

if (planName.includes(LightRailName))
///....
```

server/api/lib/tags/light_rail.js
```diff
@@ -0,0 +1,32 @@
...
+    const getReturnDict = (stringInPlanName) => {
+      return {
+          plan_id: plan.id,
+          tag_id: tagsResources.tagNameToTagId[TAG_NAME],
+          display_score: 0, /* TODO: Add the correct display score here */
+          created_by_data_rules: JSON.stringify([{ rule: `has "${stringInPlanName}" in plan name` }])
+      };
+    };
+
+    const planName = plan.attributes.PL_NAME;
```
GalGend, line 13: Do all plans have attributes?

GalGend (commented):
Nice

Left some comments

server/api/lib/tags/light_rail.js
TomerMe2: I like to write it excplicitly. It's a part of the tags interface. If the tag doesn't apply, it should return null.
I don't want this property to be abstracted and forgotten.

After being said that, the default return of JS is "undefined". I wrote null because it means that "I decided not to return nothing", instead of "I didn't know what to return, so here's undefined stuff". 

I would like to keep the null as a part of the interface, but we can move to undefined and remove the return statement if you really want to do so.

server/api/lib/tags/light_rail.js
TomerMe2: They should, at least when they come from the DB. (Bookshelf should handle this. If he can't bring attributes, he should fail in the db fetch phase)

server/api/lib/tags/index.js
TomerMe2: It's not a class, it's a function that returns the plan tagger. I'll make the export as you stated, but in my opinion it still should be called "getPlanTagger". After you invoke it, you get the plan tagger.

GalGend: Hi, it's a common js practive - you just require functions, without needing to invoke. The class in js is a function.

server/api/controller/cron.js
```diff
@@ -218,10 +219,18 @@ const updatePlanTags = async () => {
...
-		const tags = await generateTagsForPlan(planId);
+	for (const planOrder in plans.models) {
+		const plan = plans.models[planOrder];
+
+		try {
+			await PlanTag.deletePlanTags(plan.id);
+		}
+		catch(e) {
+			// if the deletion of existing tags fails, move to the next plan
+			continue;
```
GalGend, line 230: Maybe we should log such failure?

server/api/lib/tags/forests.js
```diff
@@ -0,0 +1,17 @@
+const { isTagApplyTable4FatherCategory} = require('../tags/utils');
+
+const TAG_NAME = 'יער';
+
+const doesTagApply = async (plan, tagsResources) => {
+    return await isTagApplyTable4FatherCategory(plan.id,
+        (fatherCategory) => fatherCategory.includes('יער'),
+        TAG_NAME,
+        tagsResources,
+        {rule: 'has "יער" in a father category in chart 4'}
```
GalGend, line 10: we should use the TAG_NAME as follows:
```
 {rule: `has ${TAG_NAME}  in a father category in chart 4`}
```

server/api/lib/tags/light_rail.js
GalGend: ooh ok

server/api/lib/tags/light_rail.js
GalGend: maybe return false? Up to you

server/api/lib/tags/streams.js
```diff
@@ -0,0 +1,17 @@
+const { isTagApplyTable4FatherCategory} = require('../tags/utils');
+
+const TAG_NAME = 'נחל';
+
+const doesTagApply = async (plan, tagsResources) => {
+    return await isTagApplyTable4FatherCategory(plan.id,
+        (fatherCategory) => fatherCategory.includes('נחל'),
+        TAG_NAME,
+        tagsResources,
+        {rule: 'has "נחל" in a father category in chart 4'}
```
GalGend, line 10: same syntax as before. Decalring consts once

GalGend (approved):
Approved with minor comments

PR #425, opened 2021-09-07
Digest email sender

GalGend:
Created a new cron job, that takes all *alerts* that were last notified 7 days ago.
Then, looks for all plans that were updated or created within this timeframe.
Sending them all together using a template grid that looks like this:
image

image

Last, updating the alert to now date.


server/api/service/staticmap.js
```diff
@@ -13,7 +14,7 @@ const fetchStaticMap = (lat, lon) => {
 		height: 32,
 		coord: [lon, lat]
 	};
-	map.addMarker(marker);
```
GalGend, line 16: should be reverted

server/api/service/staticmap.js
```diff
@@ -26,6 +27,36 @@ const fetchStaticMap = (lat, lon) => {
 		});
 };
 
+const drawStaticMapWithPolygon = (lat, lon, geoJSON) => {
+	const map = new StaticMaps({
+		height:600,
```
GalGend, line 32: shoud use config

server/api/service/template_email.js
```diff
@@ -0,0 +1,72 @@
...
+	 digestPlanAlert (emailPlanParams, emailAlertParams, recipient) {
+		 const email = {
+			 from: `"${this.config.from_name}" <${this.config.from_email}>`, // sender address
+			 to: 'eyal.mig@gmail.com',
+			unsubscribeLink: `${this.baseUrl}${emailAlertParams.unsubscribeLink}`,
+			 personalizations:[
+				 {
+					 'to':[{
+						 'email': 'eyal.mig@gmail.com'
+					 }
```
GalGend, line 40: should use recipient

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+
+	}
+	catch(e){
+	}
+	finally{
+	}
+
+
+	// })
+	// .then(successArray => {
```
GalGend, line 217: should update alert to be last_email_sent as now date or latest plan sent 
also, order by earliest plans

server/api/controller/cron.js
```diff
@@ -3,12 +3,13 @@ const Log = require('../lib/log');
 const iplanApi = require('../lib/iplanApi');
 const Alert = require('../model/alert');
 const Plan = require('../model/plan');
-const Email = require('../service/email');
```
GalGend, line 6: should be reverted and duplicated

server/api/model/alert.js
```diff
@@ -196,6 +197,30 @@ class Alert extends Model {
...
 		return Knex.raw(sql);
 	}
+
+	static getAlertToNotify (userOptions) {
+		const options = userOptions || {};
+		if (!options.limit) {
+			options.limit = 1;
+		}
+		return Alert.query(qb => {
+			// qb.whereRaw('last_email_sent IS NULL');
```
GalGend, line 207: add a where condition of filtering by date

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+		}
+	};
+};
+
+const sendDigestPlanningAlerts = async () => {
+	// Send emails for each user, by new plans in his area, that
+	// have been added since he last received a digest email
+	// sendPlanningAlerts(req, res, next) {id
+	Log.info('Running digest send planning alert');
+	const lastSentDifference = 150;
```
TomerMe2, line 177: I think that we should move this to the header of the file (near the imports) and put measurement units beside it in a comment.

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+	const maxAlertsToSend = 5;
```
TomerMe2, line 178: This is not used. Also, why would we send 5 alerts? Shouldn't we send only 1 email?

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+	const timeDifference = moment.duration(lastSentDifference, 'd');
+	const date = moment().subtract(timeDifference);
+
+	try {
+		const alertToNotify =  await Alert.getAlertToNotify(date, 1);
```
TomerMe2, line 183: I think it would be best to put this 1 in a const and describe it with a name.

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+			return drawStaticMapWithPolygon(
+				centroid.geometry.coordinates[1],
+				centroid.geometry.coordinates[0],
+				planGeom
+			);
+		});
+
+		const emailAlertParams = alertToEmail(alertToNotify);
+
+		const emailPlanParams = {
```
TomerMe2, line 200: Seems like a weird thing for me, it's transfering an array to a map with indices as keys? Like, it's still an array...
I think it's best keeping the array and iterating on it later rather than creating a weird array-like map.
Other than it being better SE concept in my opinion, it allows us to change the number of items that we present in a single digest with ease.

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+			firstPlan: alertPlans[2] ? planToEmail(alertPlans[2], maps[2]): {},
+			secondPlan: alertPlans[3] ? planToEmail(alertPlans[3], maps[3]): {},
+			thirdPlan: alertPlans[4] ? planToEmail(alertPlans[3], maps[4]): {},
+		// fifthPlan: planToEmail(plans[0]),
```
TomerMe2, line 204: I think we should drop this comment.

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+		};
+		
+		await DigestEmail.digestPlanAlert(emailPlanParams, emailAlertParams);	
+
+	}
+	catch(e){
+	}
+	finally{
```
TomerMe2, line 212: We don't need this finally as of now.

server/api/model/plan.js
```diff
@@ -473,9 +515,34 @@ class Plan extends Model {
...
 	}
+
+	static getPlansByGeometryThatWereUpdatedSince(geometryPolygon, date, limit) {
+		const dateString = moment(date).format('YYYY-MM-DD h:mm');
+		return Plan.query(qb => {
+			qb.where('created_at', '>', dateString);
+			const polygon = wkt.convert(geometryPolygon);
+			qb.whereRaw(`ST_Intersects(geom, ST_GeomFromText("${polygon}", 4326))`);
+		}).fetchPage({
+			pageSize: limit || 5,
```
TomerMe2, line 530: This 5 should be a const in the header of this file in my opinion.

server/api/service/template_email.js
```diff
@@ -0,0 +1,72 @@
...
+						 ...emailAlertParams
+					 }
+				 }
+			  ],
+			 template_id : this.dynamicTemplates.DigestPlanAlert,
+		 };
+		 return sgMail
+			 .send(email)
+			 .then((res) => {
+		   console.log('Email sent');
```
TomerMe2, line 53: console.log() should be changed to Log.info()

TomerMe2 (approved):
Awesome work!

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
+	const alertTitle = `תוכניות חדשות בקרבת ${alert.get('address') ||  'תחומי הענין שלך'}`;
+	return {
+		alert:{
+			title: alertTitle,
+			unsubscribeLink: `${this.baseUrl}alerts/unsubscribe/${alert.unsubsribeToken()}`
+		}
+	};
+};
+
+const sendDigestPlanningAlerts = async () => {
```
florpor, line 172: so to my understanding the logic is as follows (correct me if i'm wrong) -
1. query db for an alert that wasn't checked in X days
2. query db for Y plans that were created since the alert was last checked
3. for each plan generate static map and format other data
4. send alert-specific email to the user the alert belongs to
5. update alert last-checked field to time when this process began

the main technical issue i can think of here is that this process involves creating a static map per every plan (that will be sent) per every alert. this may strain the tile service (osm) and also our servers (rendering tiles together plus plan polygons) and might cause some issues, plus it will take a long time to run.
maybe a map cache can be used?
this will be somewhat mitigated by the fact that not every plan is sent (since only Y plans are queried).

product questions:
1. is it okay that a plan might be sent more than once to a user (if it matches more than one alert)?
2. also, is it okay that some plans will never be sent to a user (and will the plan selection be based on any plan parameters, or will it just be the Y newest plans)?

server/migrations/20210611155614_digest_mail_user.js
```diff
@@ -0,0 +1,14 @@
+
+exports.up = async function(knex) {
+
+    await knex.schema.table('alert', table => {
+        table.datetime('last_email_sent');
```
florpor, line 5: maybe set initial value when the migration is run of all existing alerts to now so a ton of emails won't be sent with old plans?

GalGend: yes i'll add it

server/api/controller/cron.js
GalGend: You got the logic right. 
As per your first question, that is possible. Yet, we're working on a smarter digest, that will include all users alerts, so plan will be sent twice.

As per the second point, to prevent that, I'll update the last_updated of the email to be the latest date of the plans, and will fetch the plans by order. That means, that if your alert has exceeded 5 plans, you'll get the top 5 , and then, in another email, the rest. 
That is because sendgrid doesn't support arrays and loops. So I leave placeholder for 5 plans, but can increase it manually up to 10 if needed

server/api/service/landUseMappers.js
```diff
@@ -0,0 +1,68 @@
...
+	} };
+
+const INCREASE_KEY = '+';
+const DECREASE_KEY = '-';
+
+const parseLandUses = (changes) => {
+	// const changes = JSON.parse(landUseString);
+	const mappedChanges = {};
+	changes[0].forEach(change => {
+		const { key: landUseKey } = find(LAND_USES, landUse=> change[3].includes(landUse.hebKey));
```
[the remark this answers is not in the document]
GalGend: Yes , i'll add. That's how to parse the area_land_change we have in the DB

server/api/service/template_email.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * Dynamic Template Email service.
+ * Compiles emails and pushes to an email web service.
+ */
+const Promise = require('bluebird');
+const Nodemailer = require('nodemailer');
+const sgMail = require('@sendgrid/mail');
```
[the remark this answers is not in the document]
GalGend: will add

server/api/service/template_email.js
```diff
@@ -0,0 +1,72 @@
...
+ 
+const Juice = require('juice');
```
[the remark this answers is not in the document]
GalGend: Just want to split the logic of proper template mails of mustache and sendgrid template ones

server/api/controller/cron.js
```diff
@@ -145,6 +146,86 @@ const sendPlanningAlerts = () => {
...
 
+const planToEmail = (plan, map) => {
+	return {
+		id: plan.get('id'),
+		title: plan.get('plan_display_name'),
+		city: plan.get('PLAN_COUNTY_NAME'),
+		text: plan.get('goals_from_mavat'),
+		status: plan.get('status'),
+		areaChange: plan.describeHousingChange(),
+		map: 'data:image/gif;base64,'+ map,
```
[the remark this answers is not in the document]
GalGend: Eventually I send the maps as attachemnets and not as bas64 ones because it didn't work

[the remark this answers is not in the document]
GalGend: will remove

server/api/service/staticmap.js
GalGend: done

server/api/service/staticmap.js
GalGend: done

server/api/controller/cron.js
GalGend: done

PR #428, opened 2021-09-14
Tags Into the API [WIP]

### LINKED ISSUE #427: Add Tags To Plan Endpoint
Add tags to the browse function in /controller/plan.js

TomerMe2:
Closes #427

### DIFF
--- a/server/api/controller/plan.js
+++ b/server/api/controller/plan.js
@@ -5,6 +5,8 @@ const Config = require('../lib/config');
 const { Knex } = require('../service/database');
 const Exception = require('../model/exception');
 const wkt = require('terraformer-wkt-parser');
+const Tag = require('../model/tag');
+const Log = require('../lib/log');
 
 const columns = [
 	'id',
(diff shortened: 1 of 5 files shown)


server/api/controller/plan.js
```diff
@@ -73,7 +76,50 @@ class PlanController extends Controller {
...
-		return super.browse(req, q);
+		return super.browse(req, q, this.afterFetch);
+	}
+
+
+	// attached tag_name into tags. The initial fetch brings only tags id, but we want the tags name.
+	async afterFetch (collection) {
+
+		// take all the unique tag ids, and get it into an array (set doesn't work in the db query later on)
+		const relevantTagIds = Array.from(new Set(collection.models.map(planModel => {
```
GalGend, line 87: Hi @TomerMe2 can you please take a look on that link? I think bookshelf allows this functionality internally. 
https://bookshelfjs.org/api.html#Model-instance-related

TomerMe2: Seems that it can work, but it will require a db query for each plan_tag instance.
The docs says that it works on a specific model and not on a collection. Do you want me to refactor into using model.related()? I can do that, but I think that it's not the correct solution for our problem.

GalGend: That's an approach that'd enbale us to fetch all relevant models at 1 query rather than in twice

Issue #483, opened 2021-12-06
Create A Docker File


TomerMe2:
This will help all the developers run on a stable system for tests.
Currently, we debug using GitHub Actions and it sucks. We wish to have a local debugging environment.

The dockerfile should have:
1. NodeJS v14.17
2. MySQL 5.7
3. Ability to connect to a debugger from VS Code and Webstorm.

yaron-idan:
Hey @TomerMe2, 
There seems to be some work that's already been done for containerizing the application [here](https://github.com/meirim-org/meirim/blob/master/server/tests/docker-compose.yml). 
Do you know if this can be leveraged to fit the use case you're suggesting?

TomerMe2:
Thanks @raznegrin :)
Can you make a PR with the relevant files into this repo?

drorgivton:
@raznegrin @TomerMe2 can you update on status here?

raznegrin:
we (@TomerMe2  and I) created a container image with all the relevant packages, RN I have to create a Dockerfile, but until that we could use the container

GalGend:
We want to achieve that using a new (thirs) EC2 instance
@raznegrin

schshmuel:
HI, @drorgivton approched me on Facebook and asked for help in DevOps issues on your project. I would like to start with dockerzing the app. Can I do this? Is this still an open issue?

TomerMe2:
Hi @schshmuel :)
It's still an open issue. A dockerfile would be awesome.
@GalGend, can you please assign @schshmuel to this issue?

schshmuel:
Good. Do you need 2 Docker files one for ther server and one for the client? Another issue should be writing k8s yamls or docker-compose. I Strongly recommend on k8s.

TomerMe2:
@schshmuel right now we run the server and the client on the same machine without any virtualization, so we have the same environment.
I think that it will be easier for now to run them both on the same container.
If you think that it's best in our situation to run the server and client on separate containers and it will not hurt stability, we can go with 2 dockerfiles.
Let me known what you think 🙂

GalGend:
Hi all,
@TomerMe2 can you please remind me on why we need to dockerize our project?

@schshmuel as per your question, we currently deploy our app directly to Ec2, so we won't use the dockers for production/stg. env.

Also, there are more devops issues open + a devops milestone, you can take a look until we understand if we keep this issue open.

schshmuel:
Hi @GalGend,

[...]

PR #496, opened 2022-01-15
DevOps - Docker-Compose #483

raznegrin:


TomerMe2:
Thanks for PR :)
I changed the database-service/Dockerfile to work with our username and password.
Things that should be done that I don't know how to do:
1. Run the following commands in the SQL dockerfile, inside the SQL shell:
```
CREATE DATABASE meirim character set UTF8 collate utf8_bin;
SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY,',''));
```
2. Make service/start.sh run:
```knex migrate:latest```
3. Forward the UI port, which is 3000 into the user's PC.
4. Forward the API port, which is 3001 into the user's PC.
5. Allow SSH connection into the service docker from localhost only, so we would have an easy time configure a remote interpeted. Make sure that this SSH port is not exposed from the machine to the internet.
6. Remove the /service/api directory, I didn't know if I should remove it or if you need examples from it.
7. Make the start.sh script to work with our server. For example, to run the server you need to cd into /server and not into /api.
8. Integrate this PR's readme into Meirim's readme so everyone would know how to run these dockers and debug using them :)
9. Test this configuration by going into localhost:3000 on a browser to see if Meirim is running locally.

Thanks a lot for your work :)

PR #500, opened 2022-01-29
updatePlanTags In Crawler Can Be Called From A Cron Job 

### LINKED ISSUE #393: Tag Maker Job
We want to seperate the creation of the tags from the crawler.
The creation of the tags requirements are as follow:

1. Cron job that runs every 24 hours.
2. Re-compute the tags of a plan if the last update time of the plan is after the last update time of the tags of this plan. 
3. Before re-computing the tags of a plan, remove all previous tags for this plan.
4. The creation of father tags are not needed as of now, but beware that we need to add that functionality later. (Make sure that it will be comfortable to add this functionality)  
5. This code should sit in the cron directory. It should resemble the existing crons in its behavior.

TomerMe2:
I made the adjustments for updatePlanTags in order for us to call it from a cron job.
It updates the tags of the necessary plans only.

Closes #393

### DIFF
--- a/server/api/controller/cron.js
+++ b/server/api/controller/cron.js
@@ -307,7 +307,6 @@ const updatePlanTags = async () => {
 	// Re-compute the tags of a plan if the last update time of the plan is after the last update time of the tags of this plan.
 	// Before re-computing the tags of a plan, remove all previous tags for this plan.
 
-	// TODO: Loop on the plans that need to be updated
 	const plans = await Plan.getPlansToTag();
 	Log.info(`Processing ${plans.models.length} plans`);
 	for (const planOrder in plans.models) {
@@ -323,10 +322,13 @@ const updatePlanTags = async () => {
 		}
 
 		const tags = await tagger(plan);
-		if (tags && tags.length > 0){
-			await PlanTag.createPlanTags(tags);
+		// call it even without tags to update the last_tags_update field
+		await PlanTag.createPlanTags(plan.id, tags);
+
+		if (tags && tags.length > 0) {
 			tagCounter++;
 		}
+
 	}
 	let end = Number(Date.now());
 	const duration = end - start;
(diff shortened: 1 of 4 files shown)


server/migrations/20220129115321_add_last_tags_update.js
```diff
@@ -0,0 +1,15 @@
+
+exports.up = function(knex) {
+    return knex.schema.table('plan', table => {
+        // some default time
+        table.dateTime('last_tags_update').notNullable().defaultTo('2000-01-01 00:00:01');
```
GalGend, line 5: why not null?

GalGend (approved):

server/api/controller/cron.js
```diff
@@ -307,7 +307,6 @@ const updatePlanTags = async () => {
 	// Re-compute the tags of a plan if the last update time of the plan is after the last update time of the tags of this plan.
 	// Before re-computing the tags of a plan, remove all previous tags for this plan.
 
-	// TODO: Loop on the plans that need to be updated
 	const plans = await Plan.getPlansToTag();
```
GalGend, line 310: please limit to 5 plans

server/api/model/plan_tag.js
```diff
@@ -47,16 +48,20 @@ class PlanTag extends Model {
...
 				}
+
 			});
+
+			const curr_time = moment().format('YYYY-MM-DD HH:mm:ss');
+			await Knex.raw(`UPDATE plan SET last_tags_update = ? WHERE id = ?`, [curr_time, plan_id]);
+
 		} catch (err) {
-			Log.error(err);
+			Log.error('had a problem saving tags for plan ' + plan_id + '\n' + err);
```
gruppin, line 64: Log.error(`problem saving tags for plan ${plan_id} `, err);

server/api/controller/cron.js
```diff
@@ -307,7 +307,6 @@ const updatePlanTags = async () => {
...
 	Log.info(`Processing ${plans.models.length} plans`);
 	for (const planOrder in plans.models) {
```
gruppin, line 312: for await (const planOrder of plans.models) {

PR #517, opened 2022-03-18
resolve gush helka polygon with open gov map api

gruppin:


GalGend (approved):
Approved, just to understand- gush + helka are already in the DB schema for the tree alert?

gruppin:
@GalGend yes, it's already there.

PR #562, opened 2022-10-28
iplan crash on files with no extension

yaelal:
also added catches so wont crash on promise

### DIFF
--- a/server/api/controller/cron.js
+++ b/server/api/controller/cron.js
@@ -228,7 +228,7 @@ const sendDigestPlanningAlerts = async () => {
 		
 		}
 		catch (e) {
-			console.log(e);
+			Log.error("error save alert", e);
 		}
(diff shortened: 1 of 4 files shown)


server/api/controller/cron.js
```diff
@@ -366,33 +366,31 @@ const fetchIplan = iPlan =>
 					) {
 						plan.set('sent', oldPlan ? 1 : 0);
 					}
-					return plan;
```
GalGend, line 369: no need to return the plan?

GalGend (approved):

GalGend:
thank you @yaelal , please sign the liscence so i can merge the pr

server/api/controller/cron.js
yaelal: no as it went to the next then which i deleted cause it was redundant

PR #577, opened 2022-11-08
feat(geocoder): support multiple Helka

yossi-eynav:


server/api/service/geocoder.js
```diff
@@ -23,35 +24,75 @@ const degreeToMeter = (lon, lat, dn, de) => {
...
+						});
+					const geoJsonRes = res.data;
+					
+					if (!geoJsonRes || !geoJsonRes['features']) {
+						Log.error('Error in geoJson:', geoJsonRes);
+						return;
+					}
+					
+					if (geoJsonRes['features'].length === 0) {
+						return null;
```
gruppin, line 70: no logs? why to separate the cases

server/api/service/geocoder.js
```diff
@@ -23,35 +24,75 @@ const degreeToMeter = (lon, lat, dn, de) => {
...
+					}
+	
+					return geoJsonRes['features'][0]
+				} catch(e) {
+					return null
```
gruppin, line 75: log

server/api/service/geocoder.js
```diff
@@ -23,35 +24,75 @@ const degreeToMeter = (lon, lat, dn, de) => {
...
+	//  - "helka 4 and 5" => [4,5]  
+	const helkaArr = [...new Set(helka.match(/\d+/gi))]
+
+
+	// filter the raw string into a single numeric value 
+	// e.g - "194,194" => 194
+	const filteredGush = (gush.match(/\d+/i) || [])[0]
+	
+	
+	console.log('filtered gush and helka', 'helkaArr', helkaArr, 'filteredGush', filteredGush)
```
gruppin, line 50: please remove or use Log.info/debug

server/api/service/geocoder.js
```diff
@@ -23,35 +24,75 @@ const degreeToMeter = (lon, lat, dn, de) => {
...
-		Log.info('Gush Helka polygon is:', JSON.stringify(geoJsonRes['features'][0]['geometry']));
-		const geometryRes = geoJsonRes['features'][0]['geometry'];
-		if (geometryRes['type'] == 'Polygon') {
-			return geometryRes;
+		const features = (await Promise.all(
+			helkaArr.map(async (helka) => {
+				try {
+					const res = await axios(
+						{
+							url: `https://open.govmap.gov.il/geoserver/opendata/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=2.0.0&TYPENAMES=opendata:PARCEL_ALL&TYPENAME=opendata:PARCEL_ALL&outputFormat=json&cql_FILTER=GUSH_NUM=${filteredGush}AND%20PARCEL=${helka}&srsName=EPSG:4326`,
```
gruppin, line 58: I know the ugly peace of code is mine, but if you could extract the url to const I would appreciate it :)

server/api/service/geocoder.js
```diff
@@ -23,35 +24,75 @@ const degreeToMeter = (lon, lat, dn, de) => {
...
 		return;
 	}
+
+
+	// parse the raw string into a unique array of numbers 
+	// e.g -
+    //  - "123,123" => [123]
+	//  - "1-2" => [1,2]
+	//  - "helka 4 and 5" => [4,5]  
+	const helkaArr = [...new Set(helka.match(/\d+/gi))]
```
gruppin, line 42: consider adding range:
"1-3" => [1,3] => [1,2,3]

gruppin (commented):
Hi! looks great!
In this pr you addressed only the case on single gush with multiple helkot.
what about multiple gushim?
this would require more complex gush/helka resolving - not to fail the Promise.all in case of missing gush/helka (since we don't know if the helka exists for each gush) - but it worth handling IMO.

server/api/service/geocoder.js
```diff
@@ -23,35 +24,75 @@ const degreeToMeter = (lon, lat, dn, de) => {
...
+
+
+	// filter the raw string into a single numeric value 
+	// e.g - "194,194" => 194
+	const filteredGush = (gush.match(/\d+/i) || [])[0]
```
gruppin, line 47: missing ;

server/api/service/geocoder.js
gruppin, line 75: missing ; in end of lines

server/api/service/geocoder.js
yossi-eynav: mmmmmm I can add a log

server/api/service/geocoder.js
yossi-eynav: I can but then we need to parse it since we build the params dynamically

server/api/service/geocoder.js
yossi-eynav: will do when we'll have a DB for this data

server/api/service/geocoder.js
yossi-eynav: we don't need it - we have another log later if the arr is blank

Issue #587, opened 2022-11-25
Tree felling data for Tel Aviv


electricmonk:
Hi folks,
I noticed that there's no data for Tel Aviv in the tree felling permit list. Is this a bug?
Looking at the codebase here: https://github.com/meirim-org/meirim/blob/f4c2d44d018f70ba931303b218d5d3706d9e9e10/server/api/lib/trees/regional_tree_permit.js

It seems like there are urls for the relevant data. Since I'm not a regular volunteer - is there a reason this doesn't work? can I help getting it to work?

gruppin:
Tel Aviv felling licences  are published here: []((https://www.tel-aviv.gov.il/Residents/Environment/Pages/Trees.aspx))
sadly, Tel Aviv municipality don't expose to html enough data in order to crawl: missing the trees data, the permit number, the activation date etc.

Tel Aviv is not the only city missing, currently most of the big cities are not covered (besides Jerusalem) due to the structure of the woodland office administrative structure

electricmonk:
Is it worthwhile to attempt scraping the PDF files?

gruppin:
for a long time my answer was no, since the ministry of agriculture people have promised us a  brand new, unified tree felling system. now i'm not sure...

yaelal:
cant we just download the csv and parse that? https://www.gov.il/he/departments/guides/pro_felling_trees

electricmonk:
> i'm probably missing something, but cant we just download the csv and parse that? https://www.gov.il/he/departments/guides/pro_felling_trees

this is the URL I was referring to

electricmonk:
`server/api/model/tree_permit_constants.js`:

```
exports.UNSUPPORTED_PLACES = [
	'אשדוד',
	'באר שבע',
	'גבעתיים',
	'גבעת שמואל',
	'הוד השרון',
	'חיפה',
	'יבנה',
	'יפו',
	'נתניה',
	'פתח תקווה',
	'ראשון לציון',
	'רחובות',
	'רמת גן',
	'תל אביב-יפו',
];
```

Why are these unsupported? @gruppin it was added here:
https://github.com/meirim-org/meirim/commit/3e53c34c093e332eadb77d193a93fd4b61b57a60

electricmonk:
I get this by running the crawler against the above URL:
```
TreePermit {
  attributes: [Object: null prototype] {
    regional_office: 'פקיד יערות מחוז תל אביב',
    permit_number: '1116',
[...]

PR #590, opened 2022-11-28
Proof of concept: parsing PDF tree licenses

electricmonk:
Hi
Following #587, here's a proof of concept that reads from the Tel Aviv city hall website, downloads PDFs and then attempts to parse them.

Currently the results are so-so; only about 90% of the PDFs are text - the rest are scanned images, which would require integrating OCR. Even those that are text, are not deterministically parsed and would require more work to improve the ability to extract reliable data.

The important question before I move forward would be - is this better than nothing? should I invest more time?


gruppin:
hi, unfortunately this is not better than nothing.
our integrity for our users is to have all the tree licenses, not only part, but they relay on us to notify about every tree license.
so in this sense, notify about partial set of licenses is even worse than not notifying at all, since we might mislead our users with a false image of reality.
I don't think you should invest more time in it.

electricmonk:
Although, we can extract 100% of street addresses from the PDF file name, and we can conceivably create a unique id from street address city and publication date. So we won't miss any petition - just have missing data for 10%-30% of them. Wdyt?

PR #595, opened 2022-12-05
feat(permits): add permit page, condition link to show if admin

Macpire:
- Nav refactoring on desktop
- adding empty page for permits
- condition new links to admin users


Macpire:
The admin condition for link I've added blindly based on the user object passing into the navigation components.
Could not register a new user and mark it admin.

client/src/pages/Permits/controller.js
```diff
@@ -0,0 +1,13 @@
+
+import api from 'services/api';
```
GalGend, line 2: Rename to PermitApi?

GalGend (approved):
looks great @adfineberg @Macpire , let me know how to proced- whether you want it merged/ deployed to stg and how we can integrate the data into it. What config should be added etc

Issue #597, opened 2022-12-06
Add Haifa trees


tomers:
Source file: http://www1.haifa.muni.il/trees/rptPirsum.pdf
Website: https://www.haifa.muni.il/development-and-construction/engineering-administration/uprooting-trees/

# Design:

This solution is based on earlier work by me to parse Haifa's tree felling permits which are published as PDF into an Excel document. The source code is available here: https://github.com/tomers/haifa-tree-felling-permits. This external project will publish a docker image for us to use. The docker image does the download, parse and store actions, and will also allows uploading to a custom S3 bucket (provided the relevenat credentials).

We will run a daily cron job to run this docker (we will always pull the latest version of the docker image), and provide it with parameters to store the output Excel file to our S3 bucket.
We will also store the file in [Athena style partitions](https://docs.aws.amazon.com/athena/latest/ug/partitions.html), so we can have a backup of out files from all our daily runs.

Now that we have the latest Excel file available as a URL, we can use existing framework to download and parse it, using a specific tree permit parsing code for Haifa.

CC @gruppin

tomers:
In order to periodically parse the public PDF from Haifa municipality's website, and store it as Excel file which is parseable from our code, we should periodically (daily) run a cron job which performs the CLI command described in https://github.com/tomers/haifa-tree-felling-permits#quick-start. 
The command requires AWS credentials, and an AWS S3 bucket with appropriate permissions which the output files will be uploaded to (e.g. `haifa-tree-felling-permits`, as in the example  in the Quick Start paragraph).

@gruppin could you please setup this, so we'll have daily Excel file created by `cron`?

GalGend:
Happily closing this one, if you have follow- up issues, please open tickets @gruppin @tomers

PR #600, opened 2022-12-08
Add support for Haifa tree felling permits (issue #597)

tomers:


GalGend:
Hi @tomers is this ready for review?

tomers:
> Hi @tomers is this ready for review?

Sure it is!
This code assumed the Excel file exist in a given URL. I will separately add the conjob task which creates the Excel file.

GalGend:
That's great! Is the file there on the address? Can you please attach here an example file?

tomers:
Yes, the file is here:
https://haifa-tree-felling-permits.s3.eu-central-1.amazonaws.com/rptPirsum.xlsx

This is a temporary bucket. We will need to establish our own.

Issue #596, opened 2022-12-05
Status scraping code fails frequently


muamichali:
**Steps to recreate**: 
 `node -e 'require("./api/controller/cron").fetchPlanStatus()'`
 
 **Results**:
 Getting an error inside fetchPlanData while getting jsonContent: Unexpected end of JSON input
 
 PS If scraping the status for one plan fails, the whole operation will halt. It should recover gracefully.

muamichali:
CC @GalGend as we discussed to fallback on scraping bee if pupetter fails. 
---
@yaelal will give me the scraping bee credentials

----------
**Scraping bee details**
2,500/month 
30 currently used
each request is $10c

PR #623, opened 2023-01-03
switch to using scrapingbee from puppeteer

muamichali:
This will address issue #596 

I tested it locally with a trial API Key because I didn't want to waste our credits on testing.

### DIFF
--- a/server/api/lib/mavat/index.js
+++ b/server/api/lib/mavat/index.js
@@ -183,22 +183,26 @@ const fetch = (planUrl, fetchPlanInstruction = true) =>
 const fetchPlanData = (planUrl) =>
 	new Promise((resolve, reject) => {
 		(async () => {
-			
-
+			const page = await browser.newPage();
+			Log.debug('Loading plan page', planUrl);
 			try {
-				Log.debug('Loading plan page', planUrl);
+				await page.goto(planUrl);
+				await page.waitForSelector('body > pre');
+				const jsonContent = await page.evaluate(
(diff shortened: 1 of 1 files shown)


server/api/lib/mavat/index.js
```diff
@@ -182,27 +183,19 @@ const fetch = (planUrl, fetchPlanInstruction = true) =>
...
 			try {
 				Log.debug('Loading plan page', planUrl);
 				try {
-					await page.goto(planUrl);
-					await page.waitForSelector('body > pre');
-					const jsonContent = await page.evaluate(
-						() => document.getElementsByTagName('pre')[0].innerText
-					);
-
+					const jsonContent = await proxy.get(planUrl);
```
GalGend, line 191: Hi, did it work? Also, we though the scrapping be could be a fallback solution, meaning we should do that on the catch, didn't we?

muamichali: Yes, it works really well!

Uh oh. You are right. I totally forgot that we said it would be a fallback. Let me fix that and also the tests.. will lyk when ready to review again.

muamichali: Ok, I changed it to use scraping bee as fallback. This also fixed the server tests. Please review again 🙏🏽

GalGend: great!!! I really wish we had this one before as large plans already failed.

GalGend (approved):

PR #630, opened 2023-01-12
Issue 565

yaelal:


server/api/lib/trees/ramat_gan_tree_permit.js
```diff
@@ -0,0 +1,131 @@
...
+                
+                const attributes = {
+                    [REGIONAL_OFFICE]: 'רמת גן',
+                    [PLACE]: 'רמת גן',
+                    [APPROVER_TITLE]: 'פקיד יערות עירוני רמת גן',
+                    [PERMIT_NUMBER]: permitNumber,
+                    [STREET]: street,
+                    [ACTION]: action,
+                    [LAST_DATE_TO_OBJECTION]: last_date_to_objection,
+                    [REASON_DETAILED]: raw['סיבת הכריתה'],
```
gruppin, line 75: should be `REASON_SHORT`

gruppin (commented):
hi! Looks great!
few comments:
1. the url to the Ramat Gan site in the app:
 should be updated. 
consider move the url logic to BE, and add a column url for the tree table.
( I managed this issue ugly with Jerusalem, perhaps this is the time to do it right ).
ugly code in here:
`meirim/client/src/pages/Tree/hooks.js`.
In ramat gan it's interesting, since they provide the actual license file, so we can link it.

2. The tree numbers crawling is not ideal, we have some misses. this is due to inconsistent input format, (some with `-`, some with ` `, some with both), but i believe it could be improved.
3. reason : should be short, not detailed.
4. missing `person_request_name`

server/api/lib/trees/ramat_gan_tree_permit.js
yaelal: didnt get any of that, should talk

yaelal: fixed REASON_SHORT and person_request_name
not sure what you want to do about the url at this point

yaelal: did what i could to improve tree numbers 
its not perfect cause there is no end to the formats they put there, but its better

yaelal: did something with the url' hope its ok

PR #639, opened 2023-01-24
#FEATURE - Links | User Story 1

microooji:


client/src/components/Links/LinkItem/LinkItem.jsx
```diff
@@ -0,0 +1,76 @@
...
+            return WebLinKIcon;
+
+        default:
+            return WebLinKIcon;
+    }
+};
+
+const getLinkActionText = (type) => {
+    switch (type) {
+        case 'whatsapp':
```
GalGend, line 34: please use the translation hook and file, `useT` and HE_IL, should be both in Hebrew and Arabic

client/src/pages/Plan/index.jsx
```diff
@@ -2,152 +2,176 @@ import React, { useState, useEffect } from 'react';
...
-						handleSubscribePanel={handleSubscribePanel} 
-						{...props}
-					/>}	
-				/>
-			</Switch>	
-		</Template>
-	);
+    const { id: planId } = useParams();
+    const [refetchComments, setRefetchComments] = useState(false);
+    // useDataHandler(planId);
```
GalGend, line 30: why is that commented out?

GalGend (approved):

client/src/pages/Plan/index.jsx
microooji: @GalGend this is just for testing without running the backend

PR #660, opened 2023-02-22
Some text changes, and cleaning up the real estate format

adfineberg:


GalGend:
@adfineberg this need to be merged right?

PR #666, opened 2023-03-08
feat(permits): permits AOI redesign

Macpire:


GalGend:
Hi @Macpire @adfineberg , this looks great, I just want to test the functionality of adding a AOI but don't have any data in the `permit_aoi` table. Can you please attach a dump/ command to add data? Thank you

GalGend:
@Macpire We need it,  but it was unused and the build fails when we have unused vars. So I commented the import and now pushed a commit with uncommenting it.

Issue #668, opened 2023-03-12
CLI tool for developers


tomers:
Have a developer one stop CLI script to perform various development tasks, such as:
- Allow/revoke the current local IP address in the security group of the server, to allow SSH to it and perform any other operations from the local development machine
- Allow for easy SSH access to remote machines (staging/production)
- Open docker such as phpMyAdmin locally, and connect them to remote DB, for easier reasoning and manipulation of our database.
-

PR #669, opened 2023-03-12
Developer's CLI tool (issue #668)

tomers:
Closes issue #668.


cli/src/phpmyadmin/commands.py
```diff
@@ -0,0 +1,65 @@
...
+    reservations = response['Reservations']
+    assert len(reservations) == 1
+    instances = reservations[0]['Instances']
+    assert len(instances) == 1
+    return instances[0]['PublicDnsName']
+
+
+@click.command()
+@click.option('-z', '--zone', type=click.Choice(ZONES, case_sensitive=False), default=DEFAULT_ZONE)
+def phpmyadmin(zone):
```
gruppin, line 49: what is php admin and for what it's used?

cli/src/cli.py
```diff
@@ -0,0 +1,28 @@
...
+
+@click.group()
+def entry_point():
+    """Manage Meirim"""
+    pass
+
+
+entry_point.add_command(authorization.authorization)
+entry_point.add_command(phpmyadmin.phpmyadmin)
+entry_point.add_command(ssh.ssh)
```
gruppin, line 25: to my understanding, the authorization and the ssh modules are suffice here.

cli/src/authorization/commands.py
```diff
@@ -0,0 +1,111 @@
...
+def my_external_ip():
+    """Get current external IP"""
+    click.echo(get_external_ip())
+
+
+@authorization.command()
+@click.option('-z', '--zone', type=click.Choice(ZONES, case_sensitive=False), default=DEFAULT_ZONE)
+@click.option('-o', '--owner', required=True, default=os.getenv('USER'), help='Owner of the added access rules')
+def authorize_my_ip(zone, owner):
+    """Authorize security groups with current external IP address"""
```
gruppin, line 103: I assume i should already be logged in into the cli right?

gruppin (commented):
looks great!! it would save sooo many time
I would try to keep this cli tool as compact as possible, i assume (perhaps mistaken) that the phpmyadmin is unnecassary here.

cli/src/authorization/commands.py
tomers: Yes, you should have your AWS CLI credentials set. Otherwise, the script will fail.
I don't know how I can automate this. Maybe we can add an interactive stage to the dev-env installation script, where it will ask you for this info. Let's open a ticket for it, and do it separately: https://github.com/meirim-org/meirim/issues/670.

cli/src/cli.py
tomers: I strongly disagree. The tool should be a one-stop tool for any development purpose.
I do agree that the hierarchy should be minimal, thus phpMyAdmin is not appropriate here, rather it should be nested under something else. However, this is a work in progress, and I didn't want to create a deeper hierarchy while I got only one item at hand, so I put it in the top level. We will definitely want to move it later on. There will be many more such tools in the future.

tomers:
As for phpMyAdmin, it gives web UI to the remote database. It allows the user to view and modify the database content. I think it is an indispensable tool.

cli/src/phpmyadmin/commands.py
tomers: I removed it now

cli/src/authorization/commands.py
tomers: > I assume i should already be logged in into the cli right?

Clarification - you are not supposed to be logged into the AWS CLI for the `meirim` CLI tool to work.
Some commands, like this one, do require AWS CLI to be configured. I added a config command to configure the AWS CLI tool:

```
> meirim config aws-cli-setup
```

tomers:
@gruppin is there is somehting that holding this PR from being merged?

Issue #692, opened 2023-05-26
Fix status scrapping- ignore bad plans


GalGend:
Currently, our status scrapper scraps plans that do not have their MP_ID. We should ignore them, as we don't have their ref on the new Mavat website.

This task is about making the status scrapper work without taking  plans that their mp_id value is NOT_FOUND

PR #693, opened 2023-05-26
not crawling on broken plans

GalGend:
1. Not fetching plans that have no MP_ID, unable to scrap them

Issue: #692


GalGend:
resolves #692

PR #739, opened 2023-07-22
Tlv trees

gruppin:


server/api/lib/trees/tlv_tree_permit.js
```diff
@@ -0,0 +1,166 @@
...
+					return null;
+				}
+
+				const treesPerPermit = parseTreesPerPermit(raw[TREE_TYPE], raw[TREE_NUM]);
+				const totalTrees = sum(treesPerPermit);
+
+				const attributes = {
+					[REGIONAL_OFFICE]: TEL_AVIV_CITY,
+					[PLACE]: TEL_AVIV_CITY,
+					[APPROVER_TITLE]: raw[APPROVER],
```
yaelal, line 94: לא יודעת אם יש מה לעשות אבל זה מכיל את השם
ערכים כמו חיים גביאל- פקיד היערות העירוני
נראה לי שזה צריך להיות ב approver_name

server/api/lib/trees/tlv_tree_permit.js
```diff
@@ -0,0 +1,166 @@
...
+					[PERMIT_NUMBER]: raw[LICENSE_NUMBER],
+					[STREET]: raw[STREET_NAME],
+					[GUSH]: raw[PERMIT_GUSH],
+					[HELKA]: raw[PERMIT_HELKA],
+					[ACTION]: SHORT_ACTION,
+					[LAST_DATE_TO_OBJECTION]: last_date_to_objection,
+					[START_DATE]: figureStartDate(null, raw[OBJECTION_TILL], HOUR_PERMIT, DATE_FORMAT_PERMIT, true),
+					[PERSON_REQUEST_NAME]: raw[LICENSE_OWNER],
+					[REASON_DETAILED]: raw[LICENSE_REASON],
```
yaelal, line 103: reason_short is not populated and we usually put something there...

yaelal:
image
ui here looks weird

server/api/lib/trees/tlv_tree_permit.js
gruppin: right, but they do not specify it, only the detailed one.
It looks bad in the app when i put it the reason in the reason_short field.

yaelal: ok

gruppin:
re: the photo. indeed, but i don't have a quick solution for it, and it looks as bad in the tlv site.
worth open a ticket for this case.
the best case of course would be that tlv muni would separate the trees by type.

server/api/lib/trees/tlv_tree_permit.js
gruppin: changed

yaelal: maybe u didnt push? i see no diff

PR #747, opened 2023-09-11
feat(ci): add CI for client side app

yossi-eynav:


yossi-eynav:
@GalGend please approve this, everything is ok from my tests.
lets deploy it and test it together next week (it won't effect anything)

client/config/default.json
```diff
@@ -1,6 +1,6 @@
 {
     "axios": {
-        "baseURL": "/api",
+        "baseURL": "https://stg.meirim.org/api",
```
GalGend, line 3: why was it changed? We use it for BE in both prod and stg. Would we want to test it before hand in production?

yossi-eynav: we must have it because we now have several envs

PR #753, opened 2023-11-27
hod hasharon tree url fix

yaelal:


yaelal:
they have one license without a url so it crashed. should work now

gruppin (approved):

